PR #90, opened 2019-01-28
Python testing

oerc0122:


TysonRayJones:
Don't accept this pull request - seems to be duplicating commits again

aniabrown:
This won't be merged into develop or master until the duplicating commits problem is solved, but needs to be merged into PythonTesting for work to go forward on that. 

Are you able to see which specific commits are being duplicated? (it isn't any of the commits in the recent history)

TysonRayJones:
Oh ok.
It seems to be all commits from the previous version until I tried to
update the doc?
I will check this thoroughly though

aniabrown:
Can you try the following command to see duplicated commit messages?
git log --oneline | awk '{$1=""; print $0;}' | sort | uniq -D
For me this gives around 70 commits with the same commit message, but different commit hashes and what seem to be legitimate differences

aniabrown:
Ok, I've found the duplicated commits. There are around 80, affecting around 200 files. 

The original commits on master are:
3bd9c93e0ad1226fe2dd1229299e9c45c451fd14 (16th May 2018) - 14af07fb652b2ea91d4d68e37635e42f0945ff20 (14 July 2018)

These commits were then merged into develop on Dec 10 and for some reason given new hashes, as:
4fea91314cc793884d3583e6948074d4dc0d1082 - 90ca2375ce29700f7c3bdfb3a2edefedec649fff

These were then merged back into master, as new commits. Master therefore has two copies,  develop has one copy of these commits. 

There are only 80 commits so this isn't a huge problem. As trying to fix this would involve changing the history of the master branch, which I imagine will cause problems for other users, I suggest we just leave the history as is and go back to pulling develop into master.

TysonRayJones:
Will this keep the duplicates? I'm fine with this provided no more get
created

aniabrown:
This will keep the existing duplicates but doesn't appear to create any new duplicates (I'm testing that by creating a pull request develop to master)

Issue #115, opened 2019-03-01
New feature: amplitude dampening


TysonRayJones:

nfwvogt:
Is that supposed to correspond to zero-temperature qubit damping, where a qubit decays with rate (or probability) p from state 1 to state 0?
Or in the Kraus language the operation that gives two Kraus operators:
[[0, sqrt(p)],
[0, 0]]
and
[[1, 0],
[0, sqrt(1-p)]]

I'm asking because we just noticed that we (@HQSquantumsimulations) need to have this kind of damping for some simulations we need to run and put an implementation of a corresponding applyOneQubitDamping into a [fork](https://github.com/HQSquantumsimulations/QuEST/tree/develop_damping) and were considering submitting a pull request.

TysonRayJones:
Yes! That's fantastic news, we'll check it out. We may be a little slow, since we'll use this opportunity to set up a protocol for reviewing and accepting external pull requests.

nfwvogt:
No worries, I'll just submit a proper pull request then. Once you have set up the protocol, just let me know what else you need.

TysonRayJones:
This has been tested, tweaked and merged into master!

PR #117, opened 2019-03-05
Add applyOneQubitDampingError function to do zero-temperature qubit damping

nfwvogt:
Pull request corresponding to #115 
The applyOneQubitDepolariseError and applyTwoQubitDepolariseError functions in QuEST provide depolarisation errors that correspond to qubits in contact with an infinite temperature bath, meaning that after full depolarisation the probability to find the qubit in a certain state is equal for the lower energy state 0 and the higher energy state 1.

To handle simulations with energy loss (or rather zero temperature) a applyOneQubitDampingError function was added bases on applyOneQubitDepolariseError. The new function only allows for qubit decay from the higher energy 1 state to the lower energy 0 state with a rate (or probability) p. (This corresponds to zero temperature according to detailed balance)

In the language of Kraus operators, the function corresponds to applying the two Kraus operators
[[0, sqrt(p)],
[0, 0]]
[[1, 0],
[0, sqrt(1-p)]]
Or in the language of superoperators to the superoperator in matrix form
[[1, 0, 0, p ],
[0, sqrt(1-p), 0, 0],
[0, 0, sqrt(1-p),0],
[0, 0, 0, 1-p]]

We are still in the process of testing the distributed and GPU versions of the function


TysonRayJones:
Thanks again for this fantastic contribution! This has caught us in the middle of migrating our unit testing to a python-based system (and potentially to CI), so we'll need a little bit of time to go over this. After a quick look over however, I want to thank you for the impressive effort in maintaining our code convention! :)

aniabrown:
Pulled into develop after making small bug fixes: 

1) typo in GPU version that was preventing compile

2) densmatr_oneQubitDephase expects argument qreal dephase, a measure of the strength of the degradation. It calculates and uses a value retain = 1-dephase in the function. The gpu and distributed versions of applyOneQubitDampingError reuse the oneQubitDephase function following the model of the oneQubitDepolarise function, but mistakenly send retain rather than dephase. The dephase and depolarise functions could use better internal commenting to make their usage clearer.

Passing unit tests on:
serial cpu: linux, gcc-6.5
gpu: linux, cuda 9, compute capability 5.2 gpu

** Not tested: distributed cpu: waiting on update to unit testing framework -- need to run gates on high qubit values to test distributed version functions

Have updated acknowledgements and AUTHORS.txt to include these contributions

PR #125, opened 2019-04-03
Python testing

oerc0122:
Completed Python testing framework. Should work with MPI/GPU integrated into CMake build system, etc.

Docs to be written in Devel/User documentation.


oerc0122:
The requirements have always been 3.4+ due to certain modules used, given python 3.4 was released  March 2014, any reasonable system will have it. Even Archer has 3.4.3.

PR #139, opened 2019-05-07
fixed bug in distributed version of controlledPauliY

aniabrown:
Maths was not quite right in the distributed version. Also needed to account for the fact that in the distributed version sometimes we are operating on upper and sometimes on lower halves of a block, and there is a sign difference in how they are updated. This is handled by passing through (-1)xconjFac in the case of lower halves of a block.


aniabrown:
Tested on linux

Single cpu: gcc 7.3, 8 threads
MPI: gcc 7.3, mvapich, 4 processes, 4 threads
GPU: gcc 6.5, CUDA 9.1, compute capability 5.2

PR #158, opened 2019-06-19
Visualstudio

BalintKoczor:
Some fixes for building on Windows (using Visual Studio).
The code compiles and runs, but I havent tested its functionality.

I use static linking in the cmake file:
elseif (WIN32)
    add_library(QuEST STATIC
        ${QuEST_SRC}
    )
Building a dynamically linked library .dll would require
a bit more effort.


TysonRayJones:
FYI I merged this patch into QuEST-MMA and it compiled successfully (well, it solved the QuEST core specific issues ;) )

TysonRayJones:
Btw it's good to give each commit a descriptive self-contained name, so that it's easy to identify the commits later once merged.

E.g.
"patching seeding for Windows"

TysonRayJones:
This is working beautifully now. Not using Visual studio on Windows, I use

```bash
cmake .. -G "Unix Makefiles"
make
```

It is a pity that this [known bug ](https://developercommunity.visualstudio.com/content/problem/303342/compiler-warnings-for-local-variables-used-only-in.html) causes `unreferenced local variable` warnings about variables used inside OpenMP regions when multithreading, but oh well.

PR #176, opened 2019-10-10
fix bug in CPU calcHilbertSchmidtDistance (Balint)

TysonRayJones:
The function calcHilbertSchmidtDistance always returned incorrect values because of a bug -- only CPU is fixed


TysonRayJones:
(no bug in GPU)

PR #182, opened 2019-10-21
included (hardware agnostic) copyStateToGPU in API

TysonRayJones:
also patched harmless but wasteful double-copy in GPU code


TysonRayJones:
This is a very simple change (merely exposing simple functions) but is somewhat urgent, so just going to accept.

PR #207, opened 2020-01-13
Example PR

TysonRayJones:
> **Do not** accept this PR

This is an example of how to contribute a new funciton to QuEST.

This new function `calcAmpSum` is relatively simple in several ways:
- the `statevec_calcAmpSum` implementation and `densmatr_calcAmpSum` implementations are identical, so only the former is defined
- the distributed CPU version merely invokes the local CPU version and performs additional reduction, so an explicit `statevec_calcAmpSumDistributed` definition is not needed.
- besides validation, there is no common hardware-agnostic processing needed, so `statevec_calcAmpSum` is defined directly in the backends, and not in `QuEST_common.c`

The complications of this function were:
- additional user-input validation (in `QuEST_validation`) needed to be defined
- in multithreaded mode, this function needed an OMP reduction
- in distributed mode, this function needed an MPI reduction
- in GPU mode, this function needed a parallel & sequential reduction

### DIFF
--- a/QuEST/include/QuEST.h
+++ b/QuEST/include/QuEST.h
@@ -3291,6 +3291,17 @@ ComplexMatrixN bindArraysToStackComplexMatrixN(
     )
 #endif
 
+/** Returns the sum of either the real or imaginary values of every amplitude in 
+ * the given \p qureg, regardless of whether \p qureg is a state-vector or density matrix.
+ *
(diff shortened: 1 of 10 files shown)


TysonRayJones:
This example is make redundant by the architectural overhaul of v4

Issue #148, opened 2019-06-08
Warnings for C99 standard when compiling GPU


TysonRayJones:
```bash
cc1plus: warning: command line option ‘-std=c99’ is valid for C/ObjC but not for C++
```

TysonRayJones:
This actually causes an error (and compilation to fail) on MacOS!

TysonRayJones:
@aniabrown are you able to take a look at this outstanding bug in the new build system

PR #208, opened 2020-01-14
Bug fix/c99 flags

aniabrown:
Prevented error seen in #148 on passing c flags through -Xcompiler. As far as I can tell these are not strictly required as the c compiler is still receiving them. Should investigate why CMAKE_C_FLAGS and not CMAKE_CXX_FLAGS are being passed through to compile a c++ file


TysonRayJones:
It seems you forked from `master` and are pulling to `develop` (bringing the unneeded/messy doc-regen). Can we correct this somehow? Otherwise thanks very much for seeing to this!

PR #229, opened 2020-03-06
Fix overflow issues in validateNumQubitsInQureg

kshitijc:
* Added `UL` type literal for calculating `numAmps` in `validateNumQubitsInQureg`. The existing implementation caused an overflow, leading to `E_DISTRIB_QUREG_TOO_SMALL` even when `2^numQubits >= numRanks`.
* Modified the `number of amplitudes` section in `createQureg` and `createDensityQureg` tests to check for up to 64 numQubits.

### DIFF
--- a/QuEST/src/QuEST_validation.c
+++ b/QuEST/src/QuEST_validation.c
@@ -280,7 +280,7 @@ void validateNumQubitsInQureg(int numQubits, int numRanks, const char* caller) {
     QuESTAssert( numQubits <= maxQubits, E_NUM_AMPS_EXCEED_TYPE, caller);
     
     // must be at least one amplitude per node
-    long unsigned int numAmps = (1<<numQubits);
+    long unsigned int numAmps = (1UL<<numQubits);
     QuESTAssert(numAmps >= numRanks, E_DISTRIB_QUREG_TOO_SMALL, caller);
 }
  
--- a/tests/test_data_structures.cpp
+++ b/tests/test_data_structures.cpp
@@ -167,7 +167,7 @@ TEST_CASE( "createDensityQureg", "[data_structures]" ) {
             /* n-qubit density matrix contains 2^(2n) amplitudes 
              * so can be spread between at most 2^(2n) ranks
              */
-            int minQb = GENERATE( range(3,maxQb) );
+            int minQb = GENERATE_COPY( range(3,maxQb) );
             env.numRanks = (int) pow(2, 2*minQb);
             int numQb = GENERATE_COPY( range(1,minQb) );
             REQUIRE_THROWS_WITH( createDensityQureg(numQb, env), Contains("Too few qubits") );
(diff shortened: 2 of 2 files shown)


TysonRayJones:
Hi Kshitij,

Well spotted, thanks very much for the patch!

I think the change to the tests however may be from a misunderstanding.
That test code:
```C++
// too few amplitudes to distribute
int minQb = GENERATE( range(2,64) );
env.numRanks = (int) pow(2, minQb);
int numQb = GENERATE_COPY( range(1,minQb) );
REQUIRE_THROWS_WITH( createQureg(numQb, env), Contains("Too few qubits") );
```
is fooling QuEST to think it is using `2^minQb` ranks, so that trying to create any less than `2^minQb` amplitudes is invalid. This is a simple algebraic check, so `minQb < 10` was already an arbitrary and unnecessarily high number. Why should this number be increased as high as `64`, and why a power-of-2 number?

A number as high as `64` will cause a crash if the validation fails, due to trying to allocate 300 exabyte state-vector, which will disrupt the testing.

kshitijc:
Yes, I wasn't too certain about these modification to the tests either. However I'll share my thinking here:
* The existing test passed in values ranging from 2 to 10 for numQb. Since these were insufficient to catch the bug in the code, I thought we should modify the tests to pass in a value which catches the overflow, which would be a value >= 32.
* Since the function we are testing uses an unsigned long type to hold the value, I thought it would be ideal to exercise the function to its limits (2^64 in this case). I think any number >= 32 would cause a crash if the validation fails, but since the purpose of the tests here is to catch issues in the main logic, it should be okay if the tests crash to surface issues with the code.

Please let me know what you think, and I'll update the PR.
Thanks!

TysonRayJones:
That's good thinking! My only reservation is that the maximum number permitted by the type may be compiler specific (which bit me just [yesterday](https://github.com/QTechTheory/QuESTlink/commit/d04896141ee03e504d3ecf7458ece08b91665360)). In theory, we could consult [limits.h](https://stackoverflow.com/questions/2053843/min-and-max-value-of-data-type-in-c), but perhaps it's easier to just check up until `maxQb` which we previously take for granted as the max anyway.

So (line [239](https://github.com/QuEST-Kit/QuEST/blob/2cc618f8989123fbb8c85818a01d1e1f5e64c7c2/tests/test_data_structures.cpp#L239))
```C++
int minQb = GENERATE( range(2,64) );
```
would become
```C++
int minQb = GENERATE_COPY( range(2, maxQb) );
```
and similarly for `createDensityQureg`.

What do you think?

kshitijc:
That's a good point. I missed noticing that. Thank you!

TysonRayJones:
According to [here](https://github.com/isaacs/github/issues/1303#issuecomment-595595284), a 'squash and merge' should attribute you with the merge. Here goes!

TysonRayJones:
Pitifully, I _can't_ attribute the squashed commits to you (grr). Can I instead ask that you make a new PR to the [_develop_](https://github.com/QuEST-Kit/QuEST/tree/develop) branch with a single "Fix overflow issues in validateNumQubitsInQureg" commit?

This allows me to update the README with an attribution in the proper way (rather than updating master). Thanks for all the help!

TysonRayJones:
(merged to dev in #230)

PR #230, opened 2020-03-09
Fix overflow issues in validateNumQubitsInQureg

kshitijc:
* Added `UL` type literal for calculating `numAmps` in `validateNumQubitsInQureg`. The existing implementation caused an overflow, leading to `E_DISTRIB_QUREG_TOO_SMALL` even when `2^numQubits >= numRanks`.
* Modified the `number of amplitudes` section in `createQureg` and `createDensityQureg` tests to check for up to 64 numQubits.


kshitijc:
Duplicate of #229

TysonRayJones:
Fantastic, thank you very much for the patch!

Issue #235, opened 2020-03-16
Add support for OpenMP 5.0


TysonRayJones:
which is the default version in `gcc-9`. This requires updating the OpenMP pragmas to explicitly specify all (formerly, assumed) shared variables.

zv-io:
In order to maintain compatibility with OpenMP 4.0, the `firstprivate` clause is used instead of adding these variables to the existing `shared` clauses.

TysonRayJones:
patched in ec9871bd57fe8c859bd68a1a5c284b2b7e9ffee7

Issue #173, opened 2019-10-08
QuEST does not compile with GCC 9.x


fcooper8472:
Following the quick start in the README, I get the following error with GCC 9.1.0 (but no errors with GCC 8.x or 7.x).

CMake output:
```
-- The C compiler identification is GNU 9.1.0
-- The CXX compiler identification is GNU 9.1.0
-- Check for working C compiler: /usr/bin/gcc-9
-- Check for working C compiler: /usr/bin/gcc-9 -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/g++-9
-- Check for working CXX compiler: /usr/bin/g++-9 -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- GPU acceleration is OFF
-- OMP acceleration is ON
-- MPI distribution is OFF
-- Found OpenMP_C: -fopenmp (found version "4.5") 
-- Found OpenMP_CXX: -fopenmp (found version "4.5") 
-- Found OpenMP: TRUE (found version "4.5")  
-- Compiling tutorial_example.c to executable demo
-- Found PythonInterp: /usr/local/bin/python3.4 (found version "3.4.9") 
-- Testing is ON
-- Configuring done
-- Generating done
-- Build files have been written to: /home/fergus/GitRepos/QuEST/build
```
Compilation error (snippet):
```
[ 10%] Building C object QuEST/CMakeFiles/QuEST.dir/src/CPU/QuEST_cpu.c.o
/home/fergus/GitRepos/QuEST/QuEST/src/CPU/QuEST_cpu.c: In function ‘densmatr_oneQubitDegradeOffDiagonal’:
/home/fergus/GitRepos/QuEST/QuEST/src/CPU/QuEST_cpu.c:53:10: error: ‘numTasks’ not specified in enclosing ‘parallel’
   53 | # pragma omp for schedule (static)
      |          ^~~
/home/fergus/GitRepos/QuEST/QuEST/src/CPU/QuEST_cpu.c:46:10: error: enclosing ‘parallel’
   46 | # pragma omp parallel \
      |          ^~~
```

TysonRayJones:
That's weird: `numTasks` is declared `const`!

```
const long long int numTasks = qureg.numAmpsPerChunk;
[...]

PR #245, opened 2020-04-06
Support GCC 9.x

zv-io:
https://github.com/zv-io/QuEST/commit/58df7d63da98c4c3a19d26b6ab6941ad91ecd51e resolves https://github.com/QuEST-Kit/QuEST/issues/235 and https://github.com/QuEST-Kit/QuEST/issues/173.
Please verify that these changes are correct before merging.
No additional cleanup/reorganization was performed.

https://github.com/zv-io/QuEST/commit/b59e9a9c1e2c4f09ef90fe515e8aea03756c5118 removes trailing whitespace (cosmetic change) from all files in the project.
Documentation files were not touched.

https://github.com/zv-io/QuEST/commit/74deea712a86b3c94aa6c14291863ee1bdb1243b removes the assumption that non-Intel compilers target x86.
Default settings may regress performance, but users should tune the flags for their target platform.


TysonRayJones:
Hi there,
Thanks very much for an OpenMP 5 patch, and sorry for the delay - competing interests meant I had to defer most of my QuEST work for a few months.

zv-io@58df7d6 looks terrific - I'll spend some time testing, but otherwise appears a welcome patch.

I disagree with the need for zv-io@b59e9a9 (and feel it "muddies the blame").

zv-io@74deea7 is an interesting debate. Since the majority of our uses run on x86, and many are physicists who would know little about compiler flags, I think it's reasonable to keep such optimisations on by default, lest they never be discovered. Users using (what is, to the physics community at large) esoteric compilers are more likely to be equipped to remove the flag.
But I sympathise with the sentiment; platform-specific defaults are a bit yuck.

TysonRayJones:
I've now checked zv-io@58df7d6 against older gcc and OpenMP versions, and it's all working beautifully.  Could I bother you to make a new PR, that:
- pulls to our [`develop`](https://github.com/QuEST-Kit/QuEST/tree/develop) branch, rather than `master`
- includes only zv-io@58df7d6

Furthermore, I'd love to add you to our list of acknowledgements. What name would you prefer to go by? (your github username is fine too)

Thanks again, this is a terrific help!

TysonRayJones:
I've gone ahead and added this to the develop branch (ec9871bd57fe8c859bd68a1a5c284b2b7e9ffee7) as part of v3.2.0 (#257)

zv-io:
> I've now checked [zv-io/QuEST@58df7d6](https://github.com/zv-io/QuEST/commit/58df7d6) against older gcc and OpenMP versions, and it's all working beautifully. Could I bother you to make a new PR, that:
> 
>     * pulls to our [`develop`](https://github.com/QuEST-Kit/QuEST/tree/develop) branch, rather than `master`
> 
>     * includes only [zv-io/QuEST@58df7d6](https://github.com/zv-io/QuEST/commit/58df7d6)

Sure, I can do this today.

> Furthermore, I'd love to add you to our list of acknowledgements. What name would you prefer to go by? (your github username is fine too)

Zach van Rijn

Thank you.

TysonRayJones:
Note I'm trying to improve this solution, since needlessly copying a shared variable to every thread (via `firstprivate`) is wasteful. However, there doesn't otherwise seem to be a method that's both OpenMP 4 and OpenMP 5 compatible (see my SO [post](https://stackoverflow.com/questions/63016641/cant-support-both-openmp-v4-and-v5-in-gcc)). Very frustrating!

Issue #129, opened 2019-04-23
Bug: build system include directories problem with cmake 3.5.1 and cuda


aniabrown:
Using cmake 3.5.1 and GPUACCELERATED=true, there is a bug where a library added with cuda_add_library won't pick up the include files in target_include_directories. See -- https://gitlab.kitware.com/cmake/cmake/issues/16571

This was patched and does work using cmake 3.10.2. I haven't tested earlier versions of cmake. 

Possible solutions are either to use include_directories rather than target_include_directories (bad practice) or to increase the cmake minimum required version

chr5tphr:
Looking at the bug report and the resulting [commit](https://gitlab.kitware.com/cmake/cmake/-/commit/7ded655f7ba82ea72a82d0555449f2df5ef38594) this seems to have been fixed earliest in version 3.7.
CMake 3.7.0 is at this point 3 years old, so it should be reasonable to increase the required cmake version.

I created a pull request at #261

TysonRayJones:
Well spotted, cheers!

PR #261, opened 2020-08-03
Update required cmake version to 3.7

chr5tphr:
This fixes the bug described in #129 in cmake in https://gitlab.kitware.com/cmake/cmake/-/issues/14201 where cuda_add_library would not pick up the include files as given by target_include_directories.
This was fixed in this [commit](https://gitlab.kitware.com/cmake/cmake/-/commit/7ded655f7ba82ea72a82d0555449f2df5ef38594) which first made it into cmake version 3.7, and therefore this fix simply increases our required cmake version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 # author: Tyson Jones (testing)
 
 # CMake initialisation.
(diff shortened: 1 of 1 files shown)


TysonRayJones:
Agreed with revising min version, thanks very much!

PR #278, opened 2021-07-01
Implementing HermitianDiagOp

Milos9304:
Implementation of Hermitian diagonal operator. The implementation was straightforward as a diagonal operator implementation already existed. This can be useful in scenarios when memory limits are being reached as we do not need to store imaginary amplitudes for the operator.

The commit also includes a function that can calculate diagonal of the Ising spin Hamiltonian, 
``void setHermitianDiagOpElemsFromIsingHamiltonian``
which can be useful when QuEST is used to simulate variational quantum algorithms.

### DIFF
--- a/QuEST/include/QuEST.h
+++ b/QuEST/include/QuEST.h
@@ -193,6 +193,30 @@ typedef struct DiagonalOp
     ComplexArray deviceOperator;
 } DiagonalOp;
 
+/** Represents a diagonal hermitian operator on the full Hilbert state of a \p Qureg.
+ *  The diagonal is constrained to be real.
+ *
+ * @ingroup type
+ * @author Tyson Jones
+ */
+typedef struct HermitianDiagOp
+{
+    //! The number of qubits this operator can act on (informing its size)
+    int numQubits;
+    //! The number of qubits stored on each distributed node
+    int numQubitsRepresented;
+    //! The number of the 2^numQubits amplitudes stored on each distributed node
+    long long int numElemsPerChunk;
+    //! The number of nodes between which the elements of this operator are split
+    int numChunks;
+    //! The position of the chunk of the operator held by this process in the full operator
+    int chunkId;
+    //! The real values of the 2^numQubits complex elements
+    qreal *diag;
+    //! A copy of the elements stored persistently on the GPU
+    ComplexArray deviceOperator;
+} HermitianDiagOp;
+
 /** Represents a system of qubits.
  * Qubits are zero-based
  *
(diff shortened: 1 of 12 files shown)


TysonRayJones:
Hi Milos,

Thanks for the contribution! Just to clarify, assuming a single state-vector of size **2^N** + **2^N** numbers (real + complex), the memory saving is a factor **1/4**:
**2^N + 2^N** (state) **+ 2^N + ~~2^N~~** (diag)
For (seemingly related) variational simulations where one may maintain many state-vectors simultaneously, this factor becomes negligible.

For even a single density matrix, the memory saving is a negligible factor **1/(2+2^(N+1))**:
**2^(2N) + 2^(2N)** (state) **+ 2^N + ~~2^N~~** (diag)

Since QuEST is memory-bandwidth bound, I don't expect to see any performance benefit (though you're welcome to test this). It doesn't seem very worthwhile to add ~700 line, 8 new API functions and a new data structure for a factor max(1/4) memory saving.

There are otherwise a few issues however with the PR:
- it pulls to `master` branch, instead of the `develop` branch
- it pulls changes in `utilities.cpp` (etc) which already exist in `develop`
- it sometimes uses indentation inconsistent with the QuEST source code (e.g. tabs instead of 4 spaces,  in `agnostic_setHermitianDiagOpElemsFromIsingHamiltonian()`)
- it documents a number of `imag` parameters which don't actually exist in the signatures. You can check for errors like this by (from a branch split from develop) running `doxygen doxyconfig/config` in the root folder and reading the log.
- some doc is totally wrong (like for `setHermitianDiagOpElemsFromIsingHamiltonian()` parameters)
- it only unit tests `applyHermitianDiagOp()`, and none of the other functions (e.g. `setHermitianDiagOp()`, `setHermitianDiagOpElemsFromIsingHamiltonian()`)
- it creates needless additional validation functions like `validateNumElemsHermitianDiagOp()`, which perform exactly as the existing `validateNumElems()`.

The addition of `setHermitianDiagOpElemsFromIsingHamiltonian()` is a great idea, and would be a welcome contribution. I note though that:
- it might as well use the existing `DiagonalOp`
- since it is populating a object of equal size to a register (and parallelises trivially), it should use hardware acceleration, and be distribution safe
- it should not throw errors and exit from within the function; validation must happen up-front in `QuEST.c`
- it needs correct documentation 
- it needs unit testing

You are very welcome to make these changes yourself (in a PR to develop). I can otherwise write the accelerated functions and testing myself, and attribute you as author of the serial prototype. Thanks very much for your efforts in any case!

Milos9304:
@TysonRayJones many thanks for your very detailed comments, indeed I am quite inexperienced contributing to opensource, so any help is appreciated. I will work through your comments and update the PR.

As you say, I do not expect to see any performance improvements, however, I am still of the opinion that 1/4 memory improvement is helpful in cases when QuEST is run on high-performance computing clusters with the aim to simulate as many qubits as the machine memory permits. The savings are then in scales of tens(or hundreds) of gigabytes and might allow us to increase the number of qubits we can practically simulate. E.g. using 64 cores machine with 16GB memory each (hence **64\*16GB=1024GB total**) and 8 bytes per amplitude (4 for real and 4 for complex), one can simulate at most 35 qubits if a DiagonalOp is used as **(8bytes+8bytes)\*(2^36)=1024GB** and obviously, 100% memory will never be available to a program. However, if HermitianDiagOp is used, **(8bytes+4bytes)\*(2^36)=768GB** and we are able to simulate 36 qubits with up to 256GB free memory that could be used for anything else. What do you think, is it worth preparing the PR?

TysonRayJones:
No problem at all.
You're right that `HermitianDiagOp` could squeeze a single additional qubit out on machines with a power-of-2 memory, when a single pure state-vector and a single diagonal operator are alive in memory. I believe this is an extraordinary use-case not worth additional bespoke functions. Note too some systems don't use power-of-2 memory, e.g. [this GPU](https://www.pny.com/nvidia-quadro-p6000) or [this node](https://lenovopress.com/lp0643.pdf) (both of which we incidentally use here at Oxford).  

If it were a significant enough use-case, a more elegant solution (without 700 lines of code duplication) could be to give `DiagonalOp` a `isHermitian` boolean field, which flags whether the `.imag` field has been allocated and should be consulted. Then, only a single additional `createHermitianDiagonalOp()` function would be needed, and the other existing functions merely include an additional `if (!op.isHermitian) // consult imag`. I still don't believe the narrow edge-case justifies this code complication.

I do think that a function to initialise a `DiagonalOp` from an all-Z Pauli Hamiltonian is worthwhile, with the changes mentioned above. I would nominate a shorter alternate suffix like `FromPauliHamil()`, which is also more general, since the fact that all-Z Pauli Hamiltonians happen to encode a type of Ising Hamiltonian (e.g. without transverse fields) is irrelevant to the operation of the created `DiagonalOp`. Documentation and validation can then constrain the `PauliHamil` to contain only `Z` operators, and mention that the resulting diagonal operator is incidentally real.

Specifically, I'd add:

- `initDiagonalOpFromPauliHamil()`, to initialise an existing diagonal operator from an existing Pauli Hamiltonian. Note use of prefix `init` over `set`, because all amplitudes of the diagonal operator are modified.
- `createDiagonalOpFromPauliHamil()` to conveniently create a new diagonal operator from an existing Pauli Hamiltonian (inferring number of qubits) which just calls the internal `createDiagonalOp()` and `initDiagonalOpFromPaulIHamil()` funcs
- `createDiagonalOpFromPauliHamilFile()` to conveniently create a new diagonal operator directly from a file compatible with [`createPauliHamilFromFile()`](https://quest-kit.github.io/QuEST-develop-doc/group__type.html#gaa77376c5e3518186eb085f37826d6008), without needing to first create (nor destroy) a `PauliHamil` instance.

Let me know what you think! 
You're always very welcome to contribute one or more particular hardware backend versions (one of OpenMP, CUDA or MPI), and I can do the remaining.

Milos9304:
Closing this PR and creating a new one into develop branch.

PR #282, opened 2021-07-04
added multiQubitNot() and multiControlledMultiQubitNot()

TysonRayJones:


TysonRayJones:
(the squashed commit was directly commited to develop, to avoid re-commiting c8a8d8f)

PR #288, opened 2021-07-22
Phase functions with shift

rrmeister:
The functions `applyParamNamedPhaseFunc()` and `applyParamNamedPhaseFuncOverrides()` now accept the new function types `SCALED_INVERSE_SHIFTED_NORM` and `SCALED_INVERSE_SHIFTED_DISTANCE`, which work in much the same way as the already implemented `SCALED_INVERSE_NORM` and `SCALED_INVERSE_DISTANCE`, but add a constant value to each of the sub-registers (or pairs of sub-registers) before calculating the norm and distance, respectively.

### DIFF
--- a/tests/test_operators.cpp
+++ b/tests/test_operators.cpp
@@ -1528,6 +1528,29 @@ TEST_CASE( "applyParamNamedPhaseFunc", "[operators]" ) {
                 applyReferenceOp(refVec, regs, totalNumQubits, diagMatr);
                 REQUIRE( areEqual(quregVec, refVec, 1E2*REAL_EPS) );
             }
+            SECTION( "SCALED_INVERSE_SHIFTED_NORM" ) {
+                
+                enum phaseFunc func = SCALED_INVERSE_SHIFTED_NORM;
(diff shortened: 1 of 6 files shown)


tests/test_operators.cpp
```diff
@@ -1528,6 +1528,29 @@ TEST_CASE( "applyParamNamedPhaseFunc", "[operators]" ) {
                 applyReferenceOp(refVec, regs, totalNumQubits, diagMatr);
                 REQUIRE( areEqual(quregVec, refVec, 1E2*REAL_EPS) );
             }
+            SECTION( "SCALED_INVERSE_SHIFTED_NORM" ) {
+                
+                enum phaseFunc func = SCALED_INVERSE_SHIFTED_NORM;
+                int numParams = 2 + numRegs;
+                qreal* params = (qreal*) malloc(numParams * sizeof(*params));
```
TysonRayJones, line 1535: use var-length array (instead of malloc) to avoid memory leak on Catch2 error handling (which would also trigger LLVM address sanitisation failure)

PR #338, opened 2022-08-09
Fix long double operations

Milos9304:
Even though qreal can be represented by long double (if PRECISION=4), the arithmetic and trigonometric operations performed cast the value to double. This is a simple fix.

### DIFF
--- a/QuEST/include/QuEST_precision.h
+++ b/QuEST/include/QuEST_precision.h
@@ -64,6 +64,19 @@
     # define REAL_EPS 1e-14
     # define REAL_SPECIFIER "%Lf"
     # define absReal(X) fabsl(X)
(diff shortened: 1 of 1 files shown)


TysonRayJones:
This is a great idea and indeed a previous worry of mine. I have however been a bit worried about overriding the precision-specific functions for the user, hence why precision-agnostic `absReal` exists (rather than a redefinition of `abs`). Perhaps these worries are moot and anyway outweighed by the damage of internal lower-precision casts. I'll think over this and come back!

TysonRayJones:
I've imagined no reasonable user scenarios where lower-precision maths would be desired in the last 9 months hehe.
Merging into develop, from where I'll test it extensively between compilers

PR #345, opened 2022-11-01
Add function to query configuration data in library bin file

HandyKurniawan:
My job is to install QuEST on two petascale HPC computer systems. In HPC contexts, there are typically several versions of the library binary code, according to the targeted hardware configuration and use intention, e.g., one for GPU and one for MPI; each of them with Float32 and Float64, etc.

With several library binary files, it becomes important to verify that the one that is dynamically linked is the one that matches the user program. (Example 1: If the `QuEST_PREC` is off between the library and user program, in the best case, there will be `SIGSEGV`. Example 2: You think you're running with GPU (and paying for the time), but really you're just running on CPU.)

We wrote a function for that, which returns the configuration data like that. 

The proposed function `getQuESTLibraryData()` is similar to the `getEnvironmentString()` function, but 
- It doesn't require parsing a string; 
- It also returns the QuEST Precision that was used for compiling the library code (similar to the undocumented `getQuEST_PREC()` function).


HandyKurniawan:
Hi @TysonRayJones,

In case maybe you missed this PR, could you please give some review or feedback regarding this PR? 

I am looking for a way to add a reviewer to this PR, but I don't have the authorization to do it.

Thank you 😄 

Best regards,
Handy Kurniawan

TysonRayJones:
Hi Handy,

Thanks very much for your contribution to QuEST, and apologies for the profane delay!
(Myself and other devs will refocus on QuEST starting from a ~month's time, after a year away)

The function you describe sounds useful and is a welcome contribution. However, while reviewing your code, I've come to the realisation that `QuESTEnv` as it currently exists is poorly designed and should be overhauled. (It needn't be passed to many functions, and its fields aren't protected in the way some existing functions seem to assume).
After a refactor, _it_ can be consulted for the runtime info you seek from the new function.
So in fact, if you're happy, I will instead refactor `QuESTEnv` myself and list yourself and Dirk Oliver Theis as co-authors of the commit, and of the revised `QuESTEnv` type.

I have still reviewed your PR below in case it's useful for you, but due to above, you needn't address anything!

  1.  I would change the nominated field `QuEST_Prec` of your `QuESTLibraryData` type to simply `precision`, so it's more inline with the other nominated fields:
  ```
CUDA
OpenMP
MPI
threads
ranks
  ```
2. Since your `QuESTLibraryData` entity is a `struct` of primitives, it can be returned directly by `getQuESTLibraryData()`, rather than the caller passing a struct pointer. For example:
  ```C
QuESTLibraryData getQuESTLibraryData(QuESTEnv env) {

    QuESTLibraryData plibdat;
    plibdat.CUDA = 0;

    ...

    return plibdat;
}

...

int main() {

    QuESTEnv env = createQuESTEnv();
    QuESTLibraryData = getQuESTLibraryData(env);

}
  ```
  3.  Your pull request includes _much more_ code than the above function; it changes 130k lines of code between 484 modified files! This is because your switch of the base branch from master to develop duplicated some QuEST commits (which we don't want on develop branch), and because your own branch ([`HandyKurniawan:master`](https://github.com/HandyKurniawan/QuEST)) has undergone many unrelated changes (build changes, copied files, and what appears like AMD experimentation).
  In general, you can avoid this via:
      - within your QuEST fork, create a _new branch_ for a particular feature you wish to contribute, from the __develop branch__ of QuEST. It can be a pain to change the base branch later.
      - commit only the minimum code changes to your new branch (no build changes specific to your environment).
      - request this branch to merge into QuEST's develop branch.
      - _do not_ make subsequent changes to your feature branch, because such commits become automatically included in your pull request.


Again, you don't need to address the above - I'll _not_ accept this PR, but I will make a new PR which overhauls `QuESTEnv` and includes you and Dirk as authors. But I hope the above helps as general Github tips! I'll leave this PR open in the meantime as a backlog reminder.

Thanks very much,
Tyson

HandyKurniawan:
Hi Tyson,

Thanks for the detailed response!

I appreciate your feedback and the thorough review of the code. It sounds like the best course of action is for you to proceed with the refactoring of QuESTEnv as you've described. It makes sense to improve the design and structure of QuESTEnv to better align with the codebase and simplify its usage.

Thank you for considering me and Dirk Oliver Theis as co-authors of the commit and the revised QuESTEnv type. I'm glad to contribute to this effort.

Regarding the current pull request, I understand that it won't be accepted due to the changes in the base branch and the unrelated modifications in my branch. I'll take note of your advice on creating feature branches from the develop branch in the future to avoid these issues.

Looking forward to the improvements in QuESTEnv!

Thanks again,
Handy Kurniawan

Issue #346, opened 2023-01-13
Failed make. CATCH_BREAK_INTO_DEBUGGER(); unrecognized instruction mnemonic, did you mean: bit, cnt, hint, ins, not?


fieldofnodes:
Just downloaded git repo and running make for first time on 

Apple Mac Book Pro M1 Max
OS: Ventura 13.1

cmake version 3.25.1
Apple clang version 14.0.0 (clang-1400.0.29.202)
Target: arm64-apple-darwin22.2.0
Thread model: posix

Full error
```
[ 44%] Built target QuEST
[ 50%] Building CXX object tests/CMakeFiles/tests.dir/main.cpp.o
In file included from /Users/jmille15/Projects/QuEST/tests/main.cpp:12:
/Users/jmille15/Projects/QuEST/tests/catch/catch.hpp:8070:13: error: unrecognized instruction mnemonic, did you mean: bit, cnt, hint, ins, not?
            CATCH_BREAK_INTO_DEBUGGER();
            ^
/Users/jmille15/Projects/QuEST/tests/catch/catch.hpp:7782:79: note: expanded from macro 'CATCH_BREAK_INTO_DEBUGGER'
    #define CATCH_BREAK_INTO_DEBUGGER() []{ if( Catch::isDebuggerActive() ) { CATCH_TRAP(); } }()
                                                                              ^
/Users/jmille15/Projects/QuEST/tests/catch/catch.hpp:7761:34: note: expanded from macro 'CATCH_TRAP'
    #define CATCH_TRAP() __asm__("int $3\n" : : ) /* NOLINT */
                                 ^
<inline asm>:1:2: note: instantiated into assembly here
        int $3
        ^
1 error generated.
make[2]: *** [tests/CMakeFiles/tests.dir/main.cpp.o] Error 1
make[1]: *** [tests/CMakeFiles/tests.dir/all] Error 2
make: *** [all] Error 2
```

Any tips?

rrmeister:
This looks like an issue with our unit testing framework Catch2 not being aware that Macs can have ARM processors now. We should probably update Catch2 to address this.
Can you confirm that this only occurs when you want to compile the unit tests, but not when working with user code?

fieldofnodes:
I try to run

```
cmake .. -DUSER_SOURCE="examples/tutorial_example.c"
```

I get

```
-- CMake version: 3.25.1
-- CMake build type: 
-- Precision is 2
-- GPU acceleration is OFF
[...]

PR #347, opened 2023-01-13
Bugfix/Catch2 update

rrmeister:
Updates Catch2 to the latest available v2.x (2.13.10) to make the unit tests aware of ARM Macs, fixing #346.
Running unit tests locally on an M1 Mac revealed an integer overflow bug in the tests, which this PR also fixes, along with some required numerical tolerance increases for some tests.


rrmeister:
Wrong target branch

PR #348, opened 2023-01-13
Bugfix/Catch2 update

rrmeister:
Updates Catch2 to the latest available v2.x (2.13.10) to make the unit tests aware of ARM Macs, fixing #346.
Running unit tests locally on an M1 Mac revealed an integer overflow bug in the tests, which this PR also fixes, along with some required numerical tolerance increases for some tests.

### DIFF
--- a/tests/catch/Catch.cmake
+++ b/tests/catch/Catch.cmake
@@ -33,6 +33,10 @@ same as the Catch name; see also ``TEST_PREFIX`` and ``TEST_SUFFIX``.
                          [TEST_SUFFIX suffix]
                          [PROPERTIES name1 value1...]
                          [TEST_LIST var]
+                         [REPORTER reporter]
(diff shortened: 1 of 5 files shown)


rrmeister:
Building tests with distribution and multithreading on, and running them locally with 4 processes on Fritz yields
```
All tests passed (677564 assertions in 131 test cases)
```

fieldofnodes:
Does this mean that with distributed and multi threading off it fails? Or did I miss that?

rrmeister:
The non-distributed unit tests are run automatically by the the CI, but distributed and GPU need to be tested separately, because github doesn't support them (but we usually only run complete tests with all configurations for merges to the master branch). Since some modifications need to be made to Catch2 to perform the distributed tests properly, I ran them anyway to make sure they work.

TysonRayJones:
Looks great, nice work!

Issue #353, opened 2023-03-25
Build static library


RIvance:
Hi,
I would like to build QuEST as a static library in Linux, but it seems like there is no option to choose whether to build a static library or a shared library. 
https://github.com/QuEST-Kit/QuEST/blob/master/QuEST/CMakeLists.txt#L293 :
```cmake
if (NOT WIN32)
    set(BUILD_SHARED_LIBS TRUE CACHE BOOL "Whether to build a dynamic library")
else ()
    set(BUILD_SHARED_LIBS FALSE CACHE BOOL "Whether to build a dynamic library")
endif ()
```

rrmeister:
Hi @RIvance!
QuEST already respects the standard flag `BUILD_SHARED_LIBS` if the user sets it. It only defaults to shared/static (depending on the OS) if `BUILD_SHARED_LIBS` is not set. So if you want to build a shared lib, you can call
```
cmake [...] -DBUILD_SHARED_LIBS=TRUE
make
```
and the build system will create a shared library. Or set it to `FALSE` to get a static library in any case. If you're using QuEST as a subproject, set the variable in the parent before adding the QuEST directory.

Hope it helps,
Cheers!

PR #354, opened 2023-03-25
Add build option `SHARED`

RIvance:
Add an option to control whether to build a static library or a shared library. #353


rrmeister:
QuEST already respects the standard flag `BUILD_SHARED_LIBS` if the user sets it. It only defaults to shared/static (depending on the OS) if `BUILD_SHARED_LIBS` is not set. So if you want to build a shared lib, you can call
```
cmake [...] -DBUILD_SHARED_LIBS=TRUE
make
```
and the build system will create a shared library. Or set it to `FALSE` to get a static library in any case. If you're using QuEST as a subproject, set the variable in the parent before adding the QuEST directory.

Having an extra option for this in `CMakeLists.txt` seems unnecessary at this point.

PR #359, opened 2023-06-15
Compile for AMD GPUs using HIP

bvillasen:
This is a lightweight modification that allows to run on an AMD GPU when GPUACCELERATED=1.
To compile for AMD GPU, the compilation of the GPU code is done with HIP instead of CUDA. To choose this option, set USE_HIP=1.

I wrote some simple instructions in the "Compile for GPUs" section of the instructions right after the instructions for Nvidia. 

Hope this is useful :)


bvillasen:
I am closing this PR. I'll resubmit with modifications made from the develop branch instead of master.

PR #380, opened 2023-09-18
Made MPI calls in exchangeStateVectors non-blocking

jjacobx:
The function `exchangeStateVectors` was modified to take advantage of non-blocking MPI communication. 

Currently, QuEST uses blocking communication calls to exchange the statevector in chunks of 2GB. On large systems, which feature high memory and a fast interconnect, this caused a noticeable slowdown. By changing the communication to non-blocking, up to 10% speedup was achieved on ARCHER2, as described in the following paper: [https://arxiv.org/abs/2308.07402](https://arxiv.org/abs/2308.07402). 

Ideally the request array would not be allocated and freed in this function, however, placing it elsewhere would require introducing MPI headers in areas of the code in which they are not currently required, for example in QuESTEnv.

### DIFF
--- a/QuEST/src/CPU/QuEST_cpu_distributed.c
+++ b/QuEST/src/CPU/QuEST_cpu_distributed.c
@@ -494,8 +494,12 @@ void densmatr_initPureState(Qureg targetQureg, Qureg copyQureg) {
 
 void exchangeStateVectors(Qureg qureg, int pairRank){
     // MPI send/receive vars
-    int TAG=100;
-    MPI_Status status;
+    int sendTag;
+    int recvTag;
+    int rank;
+    MPI_Request * requests;
+
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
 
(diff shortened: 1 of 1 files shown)


TysonRayJones:
Hi there,

Thanks for an excellent contribution! Absolutely kicking myself that the chunks were ever sent serially.
I'm running the unit tests on some beefy distributed hardware, and have confirmed the code works for few chunks.

I think the `requests` malloc is fine. The number of messages depends on the `Qureg` size, and even the extreme case of a `1 TB` RAM node exchanging a RAM-filling statevector, `requests` has only ~`2000` elements. 

Do you have a preferred name to be referred to as a QuEST contributor? (examples [here](https://github.com/QuEST-Kit/QuEST/blob/master/AUTHORS.txt))

jjacobx:
Hi Tyson, thank you for accepting my contribution! You can refer to me as Jakub Adamski.

PR #410, opened 2024-06-10
added reportQuESTEnv

TysonRayJones:
and bumped default conformance to C++17, although source remains C++14 compatible; the later standard is more likely to de-mangle printed type names in reportQuESTEnv


TysonRayJones:
Have not yet tested with GPU-acceleration, as required before merge

Issue #389, opened 2023-11-20
QuEST/CMakeLists.txt broken on arm


ashterenli:
arm does not have avx.
This is a crude workaround for gnu:
```
sed -i '261,265d' ../QuEST/CMakeLists.txt 
```

and build with e.g.

```
COMMON_FLAGS="-Wall -Wextra -Wpedantic -Wshadow -Wbad-function-cast -g -O2 -fopenmp -mcpu=neoverse-512tvb -march=armv8.4-a+sve -std=c11"

  cmake .. \
    -DCMAKE_C_COMPILER=mpicc \
    -DCMAKE_C_FLAGS="$COMMON_FLAGS" \
    -DCMAKE_CXX_COMPILER=mpicxx \
    -DCMAKE_CXX_FLAGS="$COMMON_FLAGS" \

TysonRayJones:
Hi there,

The assumptive AVX flag was raised in #259 and forgotten about (hehe). QuEST's cmake build is due to be remade, given it was originally a bit rushedly created from an existing GNUmake build, causing some very minor [issues](https://github.com/QuEST-Kit/QuEST/issues?q=is%3Aissue+is%3Aopen+CMake) since. The carefully remade cmake build will ensure all flags are arch compatible.

Relatedly, we don't invoke any explicit AVX instructions; we just encourage auto-vectorising. I'm still unsure myself whether flags `-O3` and `-march=native` together already make `-mavx` redundant, or whether including the latter introduces additional opportunities for auto-vectorising. Is it verbose overkill given openmp itself may vectorise? Would love to hear your thoughts! It's otherwise something I wish to test the performance impact of during build refactoring.

fieldofnodes:
Does AVX get flagged on non x86 arch?

TysonRayJones:
@fieldofnodes the current CMake build erroneously includes the ARX flag for non-x86 arches, as the OP reports. One can simply delete the flag from the CMake file for a workaround.

TysonRayJones:
Noting we'll address this in #402

Issue #351, opened 2023-02-06
install prefix and library naming conventions


torehl:
Hi QuEST developers,

  would it be possible to include a CMAKE_INSTALL_PREFIX section in the CMakeLists.txt file? 

  In addition, I suggest you use a similar convention often used on HPC systems for the various threaded (mt/omp/gmp), distributed (mpi), and accelerated (GPU accelerated) versions of the library:

  E.g. in $CMAKE_INSTALL_PREFIX/lib please rename the builds something like 

libQuEST.so  ->  libQuEST.so.3.5.0
libQuEST.so.3.5.0
libQuEST-mt.so.3.5.0 -> libQuEST.so.3.5.0
libQuEST-mt+mpi.so.3.5.0
libQuEST-cu118.so.3.5.0

   I ended up creating separate modules files that point to separate directories. Works as well, but probably the cleanest install.

24	/cm/shared/ex3-qc-modules/apps/QuEST/3.5.0/lib/omp
368	/cm/shared/ex3-qc-modules/apps/QuEST/3.5.0/lib/omp+mpi
1668	/cm/shared/ex3-qc-modules/apps/QuEST/3.5.0/lib/omp+cu118
2376	/cm/shared/ex3-qc-modules/apps/QuEST/3.5.0/lib
19968	/cm/shared/ex3-qc-modules/apps/QuEST/3.5.0/docs
64	/cm/shared/ex3-qc-modules/apps/QuEST/3.5.0/examples
628	/cm/shared/ex3-qc-modules/apps/QuEST/3.5.0/tests/catch
1176	/cm/shared/ex3-qc-modules/apps/QuEST/3.5.0/tests
304	/cm/shared/ex3-qc-modules/apps/QuEST/3.5.0/include
23904	/cm/shared/ex3-qc-modules/apps/QuEST/3.5.0/

    A penny for your thoughts?

Ref:
https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html

TysonRayJones:
Hi there,

Apologies for the profane delay! We currently do not intend users to `make install` QuEST because they are expected to recompile the entirety of QuEST (which takes mere seconds) when changing the deployment backend (e.g. from CPU to GPU).

You raise an interesting idea - that _all_ backends could be separately compiled and installed to distinct directories. Each would still be installed separately so that a user needn't (e.g.) have CUDA installed just to compile QuEST for distributed simulation. 

[...]

Issue #259, opened 2020-07-31
Illegal instruction when AVX instruction set is unsupported


chr5tphr:
This is related to the problem in #184

When compiling on an architecture that does not support AVX, and then running QuEST, we get an illegal instruction, specifically at:
```
(gdb) bt
#0  getQuESTDefaultSeedKey (key=0x7fffffff53b0) at .../QuEST/QuEST/src/QuEST_common.c:196
#1  0x00001554fc05284e in seedQuESTDefault () at .../QuEST/QuEST/src/CPU/QuEST_cpu_local.c:306
#2  0x00001554fc0520f7 in createQuESTEnv () at .../QuEST/QuEST/src/CPU/QuEST_cpu_local.c:177
......
(gdb) list 196
195
196         double time_in_mill =
197             (tv.tv_sec) * 1000 + (tv.tv_usec) / 1000 ; // convert tv_sec & tv_usec to millisecond
(gdb) x/i $pc
=> 0x1554fc038282 <getQuESTDefaultSeedKey+95>:  vcvtsi2sd %rax,%xmm0,%xmm0
```

The culprit instructing specifically in this case is apparently `vcvtsi2sd`

Notice the missing `avx` flag in the used cpu:
```
$ lscpu
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              24
On-line CPU(s) list: 0-23
Thread(s) per core:  2
Core(s) per socket:  6
Socket(s):           2
NUMA node(s):        2
Vendor ID:           GenuineIntel
CPU family:          6
Model:               44
Model name:          Intel(R) Xeon(R) CPU           X5675  @ 3.07GHz
Stepping:            2
CPU MHz:             3189.104
BogoMIPS:            6117.96
Virtualization:      VT-x
L1d cache:           32K
L1i cache:           32K
L2 cache:            256K
L3 cache:            12288K
NUMA node0 CPU(s):   0,2,4,6,8,10,12,14,16,18,20,22
NUMA node1 CPU(s):   1,3,5,7,9,11,13,15,17,19,21,23
[...]

Issue #154, opened 2019-06-15
Inconsistent CMake naming


TysonRayJones:
(not really a big, but a naming problem)

E.g. fields `GPUACCELERATED` vs `GPU_COMPUTE_CAPABILITY`

It would be good to shorten or at least standardise these to make them easier to remember.

TysonRayJones:
Oliver has [revised](https://github.com/QuEST-Kit/QuEST/commit/35de0e4e4b9add9cbfbfa682f73651ccc5ad19c0) the names, viewable [here](https://github.com/QuEST-Kit/QuEST/blob/35de0e4e4b9add9cbfbfa682f73651ccc5ad19c0/docs/cmake.md)

PR #430, opened 2024-07-02
v4-CMake

otbrown:
Hi @TysonRayJones,

Creating a PR to discuss and critique my work on the new build system :) I'll use this as a place to keep notes and thoughts too.

I've been through and identified the following open issues which I'll try to address, as well as my own biases:
- #351 (Didn't know you could do that through CMake, but I like it!)
- #389 and #259 - I should speak to Ali about this too. I have to say my gut feeling is "turn on -O3 a.k.a. Release and forget about it" is the optimal strategy these days, but I don't normally support such a range of environments...
- #154 Suspect this one is deprecated!

Notes:
- I've bumped from 3.7 to 3.18 (which happens to be the version available in Debian 11 apt) but consider it a placeholder until the build system is complete, at which point I'll try to figure out a real minimum. I'd be very surprised if it's below 3.13, but might be willing to forgo 3.2x features...
- I personally dislike the build in place user executable functionality that was in the old CMake, but I'll make sure it continues to be supported alongside more a standard library build/install target.
-  I will need support testing against QuESTLink/Windows things!


TysonRayJones:
Wew sounds good! I'm happy to cop the Windows testing. I'll setup CI too once the build is runnable (and probably a linter, since I've already slopped in some spaghetti).

I agree it's clumsy/overbearing/amateurish to build the user's executable, though I suspect a good chunk of QuEST's user base have physics backgrounds and lack the requisite skills to make their own build, or integrate QuEST into a bigger stack. I want to make compilation as easy as possible for that group somehow. I'm open to better ideas!

otbrown:
Can confirm that I can build and run a (very simple) example code using the available new API with and without multithreading enabled (though it doesn't deploy with MT at one qubit...), so now I can get started on the _fun_ stuff. 

Notes:
- Include paths need to be rationalised but this goes hand-in-hand with setting up the install targets, since we can use separate generators for the build and install includes.
- Once I've done that I can also reimplement the user executable functionality.
- Similarly, I can now look at #351.
- On the other hand I can't make much headway with the GPU build(s) until tests are fixed.

otbrown:
Bump to CMake 3.14 incoming for policy [CMP0082](https://cmake.org/cmake/help/latest/policy/CMP0082.html#policy:CMP0082).

TysonRayJones:
Ahh I'm sorry for the rebasing and conflicting the PR! I hadn't realised the source files were being modified too. I'm intending to rename some files (`structures` to `matrices` since it' already quite long; other new data structures will have their own files) - is it okay if I rebase these too and we sort out the conflicts later? 😅 

Btw are the relative path includes changed in b87b7f53f36e87efd497300b465168f91aac44eb essential for CMake building? I was following Kolpackov's [project structure](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html) (but preserving `src` and `include` subdirecs), and enjoyed the explicit separation (e.g. like in [`validation.cpp`](https://github.com/QuEST-Kit/QuEST/blob/v4/quest/src/core/validation.cpp)) of:

```C++
// API
#include "quest/include/modes.h"
#include "quest/include/environment.h"
#include "quest/include/qureg.h"
#include "quest/include/structures.h"

// core
#include "quest/src/core/errors.hpp"
#include "quest/src/core/bitwise.hpp"
#include "quest/src/core/memory.hpp"
#include "quest/src/core/utilities.hpp"
#include "quest/src/comm/comm_config.hpp"
#include "quest/src/cpu/cpu_config.hpp"
#include "quest/src/gpu/gpu_config.hpp"

// external library
#include <iostream>
#include <cstdlib>
#include <vector>
#include <map>
```
It was also consistent in that all paths were relative to the directory containing `quest`. 

If I understand right, with the [new includes](https://github.com/QuEST-Kit/QuEST/pull/430/commits/b87b7f53f36e87efd497300b465168f91aac44eb#diff-79e7cff07728000826dabac150c979b6013c7380940cf4a2d796804d3bf19b44), some are relative to `include/` (fair enough)
```C++
#include "types.h"
```
and the rest are relative to the source file location (e.g. this in `comm/`):
```C++
#include "../core/errors.hpp"
```
This seems less elegant and consistent - is it necessary for compilation, or offer another benefit I've missed?

otbrown:
No problem! Any conflicts should be pretty trivial to resolve, as I wouldn't expect to touch anything more than includes, so happy to just keep merging as and when.

Apologies, I'd assumed it was just an artifact of the make build 😅 

Advantages of the way I've done it:
- Install directory has a standard structure: include, libs, bins. Users just need to add `-I<installdir>/include` to build their programs using the public API.
- Relative includes don't require a `-I` flag at all at compile time, and are _slightly_ less sensitive to changes in the project structure.

However, through CMake, all things are possible 😉 
- Build and install interfaces for include directories can be set separately so I can make CMake export whatever we like as as the install include interface, and we can just tell users how to correctly write their includes relative to that.
- Setting the build includes to anything is trivial and really doesn't affect users at all 🤷‍♂️ 

So very happy to change it all back, it's just a find/replace job! It's your project, so let me know which you'd prefer.

otbrown:
Library renaming is now available -- the user can either just provide a custom string to produce `lib${LIB_NAME}.so`, or turn on  VERBOSE_LIB_NAME to append to the lib name based on configuration options. This _should_ all play nicely with the install export stuff too, though I haven't tested that.

GPU acceleration and MPI distribution are now also available, but there are two issues:
- cuQuantum: nvidia don't provide a CMake integration, and I don't _realllllly_ want to write a findcuQuantum for them, though I will if I have to. Need to ponder that one.
- ~~HIP: It looks as if the HIP version builds just by using the HIP compiler to compile the CUDA which is nice, but may present some issues for CMake. The simplest solution (for me) is to just split `ENABLE_GPU_ACCELERATION` into `ENABLE_CUDA` and `ENABLE_HIP`, but will see if there's a more creative solution!~~ Never mind, there's already a `USE_HIP` option at play in v3, so I'll just make use of that! 😁

otbrown:
@TysonRayJones is the plan to support HIP in the same way (i.e. by setting `USE_HIP`)?

otbrown:
Oh and what's the minimum required CUDA version?

TysonRayJones:
> * cuQuantum: nvidia don't provide a CMake integration, and I don't _realllllly_ want to write a findcuQuantum for them, though I will if I have to. Need to ponder that one.

Ahh annoying - it's a little unexpected that they provide GNU make examples, but it makes me feel less _out of touch_ **:^)** . Maybe they're expecting most users to build cuStateVec via [CUDA Quantum](https://nvidia.github.io/cuda-quantum/0.6.0/using/simulators.html#) (which we should expressly _not_ do). 

> @TysonRayJones is the plan to support HIP in the same way (i.e. by setting `USE_HIP`)?

Yea! But you can choose any name that makes sense / is consistent, like `COMPILE_HIP`. 

Note that the GPU code (even when _not_ compiling `cuStateVec`) will use [Thrust](https://nvidia.github.io/cccl/thrust/), because:
- it's included in the CUDA toolkit
- it avoids lots of boilerplate/custom-kernels to do reductions (e.g. compare this [lovely Thrust version](https://github.com/QuEST-Kit/QuEST/blob/d4f75f724993b4af8e43a796e3c09ce24ae11670/QuEST/src/GPU/QuEST_cuQuantum.cu#L1154) of this [custom kernel](https://github.com/QuEST-Kit/QuEST/blob/d4f75f724993b4af8e43a796e3c09ce24ae11670/QuEST/src/GPU/QuEST_gpu.cu#L1663) and [caller](https://github.com/QuEST-Kit/QuEST/blob/d4f75f724993b4af8e43a796e3c09ce24ae11670/QuEST/src/GPU/QuEST_gpu.cu#L1767)).

This requires that HIP can wrap/map the Thrust functors and calls to [rocThrust](https://github.com/ROCm/rocThrust), but I don't know if that's always included with ROCm. I've just been hoping and praying! In any case, feel free to defer HIP support for the moment in case it causes a problem!

otbrown:
> most users to build cuStateVec via [CUDA Quantum](https://nvidia.github.io/cuda-quantum/0.6.0/using/simulators.html#) (which we should expressly not do)

Almost certainly! I will probably just relent and write a simple findCuQuantum. This works fine as long as standard installations of it have a consistent directory structure.

 > you can choose any name that makes sense / is consistent, like `COMPILE_HIP`
 
 For the compiler flags `COMPILE_CUDA` still makes sense as what's written _is_ CUDA whether we compile it with nvcc or hipcc, and there's no point changing the pragmas. On the other hand I may ditch `ENABLE_GPU_ACCELERATION` in favour of `ENABLE_CUDA` and `ENABLE_HIP` a) because I don't see a scenario where someone tries to build both, and if they do they have bigger problems than the CMake, and b) both are shorter than `ENABLE_GPU_ACCELERATION`. 😃 

> Note that the GPU code (even when not compiling cuStateVec) should use [Thrust](https://nvidia.github.io/cccl/thrust/),

Good to know! Seems like both Thrust and rocThrust export CMake targets, so should at least be able to make this an explicit dependency. I'll look into it (no point adding it if it breaks builds that actually _should_ work...). 

> feel free to defer HIP support 

I actually looked into it already on the ARCHER2 GPU testbed, and enabling HIP support is super easy with CMake 3.21+ (for context, the latest is 3.30, so we're still well behind the curve). Shouldn't be an issue therefore +/- potential Thrust issues. I can at least get HIP support in and building for now, and then see if it actually _runs_ in the testing/benchmarking phase.

Perfect, thanks for the versions, I'll add them in!

otbrown:
Eurgh, so although Thrust _has_ a CMake target, but for some reason CUDAToolkit [doesn't expose it](https://gitlab.kitware.com/cmake/cmake/-/issues/22891) 🙄 Best option for now to make this an explicit dependency is probably to figure out what version of CUDAToolkit bundled the minimum version of Thrust and then set that as a minimum version of CUDAToolkit... We could insist on explicitly linking Thrust, but I feel it's not worth getting the user to dig through NVIDIA install directories when the code would actually link fine.

otbrown:
Hi @TysonRayJones!

Hope you're well, and have found some time to enjoy a holiday in amongst the development work...

I've marked this as ready for review now, as I _think_ what we have is enough to be useful for your planned 'developer' release, but naturally it's your call whether you'd like to merge now or later. Note the TODOs listed below. I've added [docs/cmake.md](https://github.com/otbrown/QuEST/blob/v4/docs/cmake.md) to list the new CMake options.

I've got a paper to review, PhD thesis corrections to approve, and I'm getting married next week, so I probably won't be looking at this again before I'm back in the office 😅 

Happy new year! 🎊 

CMake TODO:
- Implement CuQuantum support. The CMake config variable exists but doesn't really do anything currently. As mentioned before this may require me to write a FindCuQuantum which should simple enough, but once I'm back in the office! This is the most egregious CMake specific issue now.
- What version of clang are you using? I stil get compilation failures on the tests with clang 16 and clang 18 I'm afraid. So I still need to test the tests, but all the required CMake is there I think.
- Test HIP support.
- Test compilation on Windows.
- Update compilation documentation, dependent on the tutorial program.

TysonRayJones:
Wew amazing! Happy holidays to you, and a massive congratulations on your wedding!!! 🎉 🎉  

I've got two items (🤞) left on my backlog before I'll start using your CMake build and test it on as many platforms as I can access. 

To your dot points:
1. I recall having trouble linking cuStateVec with CMake in QuEST v3 because I was trying to support the despairingly old default CMake version on the Oxford cluster (as I lament [here](https://github.com/NVIDIA/cuQuantum/discussions/85)). I'm happy for us to force a modern CMake minimum version if it helps. 
2. Oh my bad! I've been using only Clang 15 to run the v3 tests, i.e. test the [deprecation API](https://github.com/QuEST-Kit/QuEST/blob/v4/quest/include/deprecated.h). It is probably that optional hacky header causing the problems - I'll check it out!
3. I'll have a go at compiling with HIP and on Windows, and can also update the revised tutorial doc - your [args page](https://github.com/otbrown/QuEST/blob/v4/docs/cmake.md) looks brilliant!

Thanks very much for all your fantastic work so far! 🙌

TysonRayJones:
Howdy! I've _finally_ had time to use the CMake build. I get it to compile by... 

1. copying over the [`tests`](https://github.com/QuEST-Kit/QuEST/tree/scratch-v3-tests/tests) folder from the `scratch-v3-tests` branch (which has some patches)
2. retaining your `CMakeLists.txt` from this PR (otherwise overwritten by above) 
3. replacing the reference to `DEAULT_VALIDATION_EPSILON` (forgive the misspelling eep) in [line 38](https://github.com/QuEST-Kit/QuEST/blob/17378fb06e5ff20e382ad7a7df124ef70b3d4bbb/tests/test_utilities.hpp#L38) of `test_utilities.hpp` with a constant `1E-10`.

Of course I'll fix these shenanigans!

Disabling the tests, I can indeed compile (on my Macbook) with both `clang 15.0.0` and `gcc 14.2.0` :tada:

The tests do not compile (except in clang 15) due to [`deprecated.h`](https://github.com/otbrown/QuEST/blob/v4/quest/include/deprecated.h)'s use of [`_Generic`](https://en.cppreference.com/w/c/language/generic), like [here](https://github.com/QuEST-Kit/QuEST/blob/17378fb06e5ff20e382ad7a7df124ef70b3d4bbb/quest/include/deprecated.h#L621). I've soberingly realised that this `C11` syntax is _not_ adopted by `C++17` 😢 (test [here](https://godbolt.org/z/Pe6vq3vM8)). Only a handful of compilers (like the Frankenstein's monster that is Clang) will accept it as an extension. Indeed with multithreading disabled (see below), I can successfully compile with `clang 15.0.0`, though as you've noted, even that may go away with newer clang versions! 

So, let's just give up on cross-platform supporting the v3 unit tests, which seek only to test the deprecation API anyway. We'll instead target new, v4, C++17 compatible unit tests, which I'll prepare. The deprecation tests can run on old clang offline, since they're inessential ¯\\_(ツ)_/¯

Some other notes:
- The doc suggests `cmake -Bbuild .` from the root dir, in lieu of the standard(?) `cmake ..` from _within_ `build`. Not a problem, though we should note users should subsequently call `make -C build` to avoid confusion.
- Some additional [bull shittery](https://mac.r-project.org/openmp/) is needed to get OpenMP working with Clang compilers, shown [here](https://gist.github.com/scivision/16c2ca1dc250f54d34f1a1a35596f4a0). We can direct Clang users to install `libomp` in the doc.

I channeled my `_Generic` frustrations inspired this meme:

img

TysonRayJones:
I've scaffolded the v4 unit tests [here](https://github.com/QuEST-Kit/QuEST/tree/v4/tests) ([commit](https://github.com/QuEST-Kit/QuEST/commit/47e542a11bd2609345f7f2718f75df21ea3720e9)) which is compiling on all my tested systems with the hacky [bash script](https://github.com/QuEST-Kit/QuEST/blob/v4/scripts/compile.sh). No testing file needs an MPI compiler nor special preprocessors now :tada: The tests temporarily use the super old Catch2 single-header, which we can trivially change to modern Catch2 when supported by the CMake build (replacing these [stubs](https://github.com/QuEST-Kit/QuEST/blob/47e542a11bd2609345f7f2718f75df21ea3720e9/tests/unit/calculations.cpp#L16)).

I'll hold off on developing the new tests until the CMake build is working, because use of the bash script is nausea-inducing. Should we try to merge your build into v4 (I suppose from a fresh fork)? Totally fine if unpolished; e.g. cuStateVec or Windows aren't working, which we can sort out in another commit (and defer HIP to v4.1 if necessary). I want to experiment with using CMake to activate integration tests based on what backends got compiled (without spilling more preprocessors everywhere) 🙌

TysonRayJones:
Oh! I brainfarted hard! The deprecation header contains only a _single_ invocation of `_Generic` - the remaining uses I was remembering are actually in the v4 API, like [`matrices.h`](https://github.com/QuEST-Kit/QuEST/blob/404dbc3787ccbae187cb3cd57e3e136b9b0ce634/quest/include/matrices.h#L316), which are guarded against C++ so work fine. Ergo I can _easily_ tweak the deprecated v3 tests to work beyond clang :tada: I'll get to it

TysonRayJones:
Compiling and (well, non-distributed CPU) passing [here](https://github.com/QuEST-Kit/QuEST/pull/528) with g++ 11 on Ubuntu :tada:

TysonRayJones:
> Nice! I've put together a CMake build for the new test structure for you. As ever it runs fine on my machine, but let me know if you encounter any issues! Would be happy to merge now and either leave this PR open, or I can open a new one to resolve remaining CMake issues.

Wew brilliant! There are some minor edits I'd nominate which I'll point out using the Github review system. Up to you how to continue edits after the squash-merge!

> Managed to get through 3 papers while we had an electrician in earlier, only 10 to go...

Joy :) I'll be procrastinating mine right up until Friday it seems eheuheu

CMakeLists.txt
```diff
@@ -0,0 +1,331 @@
+# author : Ania Brown
+# author : Jacob Wilkins
+# author : Balint Koczor (Windows compatibility)
+# author : Tyson Jones (testing)
+# author : Oliver Thomson Brown (v4)
```
TysonRayJones, line 5: Feel free to move your name to the top! I suppose noting `v4` is unnecessary since it will likely stay the build for future versions. E.g.
```
# author : Oliver Thomson Brown
# contributions to previous builds from:
# author : Ania Brown
# author : Jacob Wilkins
# author : Balint Koczor (Windows compatibility)
# author : Tyson Jones (testing)
```
Because this comment doesn't get parsed by doxygen, you can use any format you wish to make it prettier, e.g.
```
# authors : 
#   - Oliver Thomson Brown
#
# contributions to previous builds from:
#   - Ania Brown
#   - Jacob Wilkins
#   - Balint Koczor
#   - Tyson Jones
```

CMakeLists.txt
```diff
@@ -0,0 +1,331 @@
...
+  string(CONCAT LIB_NAME ${LIB_NAME} "-fp${FLOAT_PRECISION}")
+endif()
+
+# Examples
+option(
+  BUILD_EXAMPLES
+  "Whether the example programs will be built alongside the QuEST library. Turned ON by default."
+  ON
+)
+message(STATUS "Examples are turned ${BUILD_EXAMPLES}. Set BUILD_EXAMPLES to modify.")
```
TysonRayJones, line 80: Do we have strong opinions on whether compiling of examples and tests should be `ON` by default? I like the default compilation (= the user's first experience) to compile the _minimum_ possible, so that:
- the compilation is faster
- the console output is readable
- the chance for exotic platform-specific warnings or compilation errors is reduced
- the user has a stronger understanding of what's going on

I'd propose the default compilation to build a _single_ C++ example file, which we could place in like `examples/tutorials/main.cpp`, and which compiles to put executable `main` in the root dir. A placeholder for now could be
```
#include "quest.h"

int main() {
    initQuESTEnv();
    reportQuESTEnv();

    Qureg qureg = createQureg(20);
    reportQuregParams(qureg);

    initPlusState(qureg);
    reportQureg(qureg);

    destroyQureg(qureg);
    finalizeQuESTEnv();
}
```
What'cha think?

docs/cmake.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+## QuEST CMake variables
+
+| Variable | (Default) Values | Notes |
+| -------- | ---------------- | ----- |
+| `LIB_NAME` | (`QuEST`), String | The QuEST library will be named `lib${LIB_NAME}.so`. Can be used to differentiate multiple versions of QuEST which have been compiled. |
+| `VERBOSE_LIB_NAME` | (`OFF`), `ON` | When turned on `LIB_NAME` will be modified according to the other configuration options chosen. For example compiling QuEST with multithreading, distribution, and double precision with `VERBOSE_LIB_NAME` turned on creates `libQuEST-fp2+mt+mpi.so`. |
+| `FLOAT_PRECISION` | (`2`), `1`, `4` | Determines which floating-point precision QuEST will use: double, single, or quad. *Note: Quad precision is not supported when also compiling for GPU.* |
+| `BUILD_EXAMPLES` | (`ON`), `OFF` | Determines whether the example programs will be built alongside QuEST. |
```
TysonRayJones, line 23: (just noting these doc'd default options would need to be change if we do end up changing the build defaults)

docs/cmake.md
```diff
@@ -0,0 +1,43 @@
...
+| `ENABLE_TESTING` | (`ON`), `OFF` | Determines whether Catch2 tests will be built alongisde QuEST. If built, tests can be run from the build directory with `make test`. |
+| `ENABLE_MULTITHREADING` | (`ON`), OFF | Determines whether QuEST will be built with support for parallelisation with OpenMP. |
+| `ENABLE_DISTRIBUTION` | (`OFF`), ON | Determines whether QuEST will be built with support for parallelisation with MPI. |
+| `ENABLE_CUDA` | (`OFF`), `ON` | Determines whether QuEST will be built with support for NVIDIA GPU acceleration. If turned on, `CMAKE_CUDA_ARCHITECTURES` should probably also be set. |
+| `ENABLE_CUQUANTUM` | (`OFF`), `ON` | Determines whether QuEST will make use of the NVIDIA CuQuantum library. Cannot be turned on if `ENABLE_CUDA` is off. |
+| `ENABLE_HIP` | (`OFF`), `ON` | Determines whether QuEST will be built with support for AMD GPU acceleration. If turned on, `CMAKE_HIP_ARCHITECTURES` should probably also be set. |
+| `ENABLE_DEPRECATION` | (`OFF`), `ON` | Determines whether QuEST will be built with support for the deprecated (v3) API. *Note: will generate compiler warnings, and not supported by GCC.` |
```
TysonRayJones, line 30: `ENABLE_DEPRECATION` might not be super clear - one could imagine it "enables deprecation of old functions, switching to user the new ones" so ergo has the inverse of the intended meaning. 

How about `ENABLE_DEPRECATED_API`? That makes it more explicit too that only the interface is changing, while the backend code is unchanged.

docs/cmake.md
```diff
@@ -0,0 +1,43 @@
...
+| `CMAKE_CXX_COMPILER` | The C++ compiler that will be used to compile QuEST. | [CMAKE_\<LANG\>_COMPILER](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER.html) |
+| `CMAKE_C_COMPILER` | The C compiler that will be used to compile QuEST. | [CMAKE_\<LANG\>_COMPILER](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER.html) |
+| `CMAKE_INSTALL_PREFIX` | The directory to which QuEST will be installed when `make install` is invoked. A standard GNU directory structure (lib, bin, include) will be used inside the prefix directory. | [CMAKE_INSTALL_PREFIX](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html) <br> [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) |
+| `CMAKE_CUDA_ARCHITECTURES` | Used to set the value of `arch` when compiling for NVIDIA GPU. | [CMAKE_CUDA_ARCHITECTURES](https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_ARCHITECTURES.html) |
```
TysonRayJones, line 42: Might be worth adding:
```
This is also known as the target GPU's "compute capability" and can be discovered [here](https://developer.nvidia.com/cuda-gpus)
```

examples/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+add_subdirectory(krausmaps)
+add_subdirectory(matrices)
+add_subdirectory(numbers)
+add_subdirectory(paulis)
+add_subdirectory(reporters)
+add_subdirectory(superoperators)
```
TysonRayJones, line 6: if we do add a `examples/tutorials/main.cpp` file, then above should include:
```
add_subdirectory(tutorials)
```

examples/paulis/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+  DESTINATION ${CMAKE_INSTALL_BINDIR}/examples/paulis
+)
+set_target_properties(paulis_cpp_initialisation 
+  PROPERTIES 
+    INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/quest/${CMAKE_INSTALL_LIBDIR}"
+    OUTPUT_NAME "cpp_initialisation"
+)
+
+add_executable(paulis_c_initialisation
+  initialisation.cpp
```
TysonRayJones, line 18: typo here; should be `initialisation.c` (rather than `.cpp`)

examples/reporters/CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
+add_executable(reporters_cpp_env
```
TysonRayJones, line 1: I apologise for all the boilerplate this requires! 😛 One day we could script/automate this (although it's not worth the time right now)

examples/superoperators/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+  DESTINATION ${CMAKE_INSTALL_BINDIR}/examples/superoperators
+)
+set_target_properties(superoperators_cpp_initialisation 
+  PROPERTIES 
+    INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/quest/${CMAKE_INSTALL_LIBDIR}"
+    OUTPUT_NAME "cpp_initialisation"
+)
+
+add_executable(superoperators_c_initialisation
+  initialisation.cpp
```
TysonRayJones, line 18: typo here; should be `initialisation.c` (rather than `.cpp`)

quest/include/precision.h
```diff
@@ -5,7 +5,7 @@
 #ifndef PRECISION_H
 #define PRECISION_H
 
-#include "quest/include/modes.h"
+#include "modes.h"
```
TysonRayJones, line 8: I'm still convinced we should follow [Kolpackov's paths](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html) (ranted about [here](https://github.com/QuEST-Kit/QuEST/pull/430#issuecomment-2220524094)), in which case all the changes to `#include "..."` of this PR would be redundant.

quest/src/api/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+target_sources(QuEST
+  PRIVATE
+  calculations.cpp
+  channels.cpp
+  debug.cpp
+  decoherence.cpp
+  environment.cpp
```
TysonRayJones, line 7: random observation - `environment` in `environment.cpp` is being highlighted by Github as if it were a CMake keyword. Is this the case?? Should the filename be avoided? 😭

quest/src/core/errors.cpp
```diff
@@ -48,7 +48,10 @@ void raiseInternalError(string errorMsg) {
 void error_functionNotImplemented(const char* caller) {
 
     string name = caller;
-    raiseInternalError("The function '" + name + "' has not yet been implemented.");
+    
+    // DEBUG
```
TysonRayJones, line 52: oops this whole function is no longer called :tada: Nothing to edit here though, I'll delete it on a subsequent commit

quest/src/core/localiser.cpp
```diff
@@ -1474,6 +1474,13 @@ void localiser_densmatr_twoQubitDepolarising(Qureg qureg, int qubit1, int qubit2
     bool comm1 = doesChannelRequireComm(qureg, qubit1);
     bool comm2 = doesChannelRequireComm(qureg, qubit2);
 
+
+
+    // TODO:
```
TysonRayJones, line 1479: hehe this comment is outdated and now untrue - banish it!

otbrown:
After a bit of cajoling (i.e. making INCLUDE_DEPRECATED_FUNCTIONS a private compiler flag for the library, and realising that test_utilities were actually needed 😅), I now have a working CMake build of the deprecated tests too 🥳 

I need to tidy it up, and probably make all compiler flags library private because that's probably good practise anyway (NOTE TO SELF: Add something to documentation so that users know that any QuEST flags are _not_ inherited by their application), but should be able to push it up later this afternoon!

scripts/compile.sh
TysonRayJones, line 1: I think I want to keep this file in the `scripts` folder - it's useful for testing build flags and seeing precisely what gets passed to the compilers

tests/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+add_subdirectory(catch)
```
TysonRayJones, line 1: ah I was hoping to keep the very-outdated `catch` folder _inside_ `tests/deprecated`, and use CMake to obtain the current version of Catch2 for the v4 tests, as discussed [here](https://github.com/QuEST-Kit/QuEST/issues/514). 

I can wrangle with this myself though if it proves a fuss - in that case, let's keep `catch` inside `tests/deprecated/` and replace the above with `add_subdirectory(deprecated/catch)` so it still builds for now

tests/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+
+add_executable(tests
```
TysonRayJones, line 3: only the deprecated tests should be seeing `test_main.cpp` and `test_utilities`, so I think they shouldn't be referenced outside of `tests/deprecated/`

tests/catch/Catch.cmake
TysonRayJones, line 1: intending not to move these files and obtain Catch2 (for v4 tests) via Cmake (as above comment)

tests/deprecated/test_calculations.cpp
```diff
@@ -120,6 +120,11 @@ TEST_CASE( "calcDensityInnerProduct", "[calculations]" ) {
  * @author Tyson Jones 
  */
 TEST_CASE( "calcExpecDiagonalOp", "[calculations]" ) {
+
+    // TODO:
```
TysonRayJones, line 124: this comment/diff is no longer true/needed, banish it!

tests/deprecated/test_calculations.cpp
```diff
@@ -480,6 +485,11 @@ TEST_CASE( "calcExpecPauliProd", "[calculations]" ) {
  * @author Tyson Jones 
  */
 TEST_CASE( "calcExpecPauliSum", "[calculations]" ) {
+
```
TysonRayJones, line 488: to the void ye scurvy outdated comment/call

tests/test_main.cpp
TysonRayJones, line 1: This file should stay as `main.cpp` since I've dropped the `test_` prefix convention of the v3 tests for the v4 ones. Happy to debate this though!

tests/test_utilities.cpp
TysonRayJones, line 1: this file not needed! All utils are in `utils/` now

tests/test_utilities.hpp
TysonRayJones, line 1: this file not needed! All utils are in `utils/` now

TysonRayJones (changes requested):
Wew lookin good! Some minor changes proposed

CMakeLists.txt
otbrown: Done! As per your second suggestion

docs/cmake.md
otbrown: That works for me

otbrown: Ha, in fact trying to find/replace this reveals that it's _actually_ `ENABLE_DEPRECATED` in the CMake😅 will update to `ENABLE_DEPRECATED_API` everywhere...

examples/superoperators/CMakeLists.txt
otbrown: Good spot, fixed!

scripts/compile.sh
otbrown: No problem, it's back!

tests/CMakeLists.txt
otbrown: Ah so I would rather not run two different versions of Catch2 at the same time, which is what we would be doing here with the deprecated API tests running Catch2 v2.7 and the new tests running Catch2 v3.x, and potentially both sitting in the same folder if the user doesn't already have Catch2 v3.x installed somewhere findable. 

Looking at your comment on #514 I see that you wanted to use some of the new v3.x functionality, so perhaps it makes sense to just bump everything to 3.x now and do away with the headers? From experience I think it should be easy enough, but I'll get through the rest of these comments before I try it!

examples/reporters/CMakeLists.txt
otbrown: At least this sort of thing really only needs to be written out once unless you make a lot of structural changes!

tests/test_main.cpp
otbrown: I do remember this being `main.cpp` at some point, so probably just careless merging on my part!

quest/src/core/errors.cpp
otbrown: Noted!

quest/include/precision.h
otbrown: Oops, I actually thought I'd changed these back, but evidently hallucinated it 😅 Fixing now!

tests/CMakeLists.txt
otbrown: Working on it on a separate branch [here](https://github.com/otbrown/QuEST/tree/catch2-v3). Mostly plain sailing except some _weirdness_ with the Generators, which I'll look into more tomorrow. Have most of a fix already though, so won't be a blocker!

TysonRayJones:
Haven't finished looking over - I was racing to merge #528 so I could merge this build PR when I was derailed by [this clang bug](https://github.com/QuEST-Kit/QuEST/issues/529). I have to tird-up _nine_ `cpu_subroutines.cpp` functions to workaround it 🤢  Do you have any thoughts about it or ideas for other workarounds?

tests/CMakeLists.txt
TysonRayJones: True, definitely gross to mix versions of Catch yea - I was envisioning migrating the old tests to new Catch was going to be an undue fuss, but brilliant if it works! No surprise the generator design has changed with Catch2 v3 hence my [lazy stubs](https://github.com/QuEST-Kit/QuEST/blob/334a128ad9420fcf1cbfb9387782f13d9bf4f630/tests/utils/lists.cpp#L38) hehe.

otbrown:
Ha, fair enough! I just got to my desk and was about to reproduce your MWE, but I'll get this Catch2 v3 tidied up instead! What's baffling to me is that as far as I can tell the code that doesn't compile matches the [example code from Catch2](https://github.com/catchorg/Catch2/blob/devel/examples/300-Gen-OwnGenerator.cpp) almost exactly, and the alternative example code which _does_ compile should do basically the same thing. In any case, if it works, I'll take the easy win...

Might still fall over at runtime, so watch this space 😆

otbrown:
It did _not_ fall over at runtime 🎉 

Will tidy it up, check _all_ the deprecated API tests run under Catch2 v3, add the bit that downloads and builds it for the user, and then merge into this branch!

otbrown:
This branch should now resolve #514 👌

tests/CMakeLists.txt
otbrown: Done!

examples/CMakeLists.txt
otbrown: It doesn't because you wanted the executable built in the root directory, and I can't add an executable with no source files even though `target_sources` exists

docs/cmake.md
otbrown: Done!

otbrown:
I've completely arbitrarily pegged it to Catch2 v3.1.1, but that can be trivially changed to anything really

TysonRayJones:
Wew absolutely brilliant!!! :tada: ❤️  🙌  Deprecated tests compile and run beautifully* on:
- serial MacOS
- distributed MacOS
- serial Ubuntu
- multithreaded Ubuntu
- multithreaded + distributed Ubuntu

I've never seen so much green text in my life!

Deprecated tests also compile fine on: 
- multithreaded + distribution + CUDA Ubuntu

although outstanding v4 GPU bugs mean they fail - that's next on my todo list ;)

The v4 test stubs of course also compile and pass on all systems 🙏 

The only remaining issues are:
- compiling OpenMP with MacOS doesn't work out-of-the-box (OpenMP not found)
- cuQuantum isn't in the header path, so cuQuantum compilation fails (unable to locate `custatevec.h`). As a quick fix, we can merely pass along [`$CUQUANTUM_ROOT`](https://docs.nvidia.com/cuda/cuquantum/24.11.0/getting-started/index.html#setting-cuquantum-root) (an env-var which even the [cuStateVec demos](https://github.com/NVIDIA/cuQuantum/blob/c202ee7ad23ce9a5072a32b45d36e8310a20eeb5/samples/custatevec/Makefile#L22) use). This makes me sad, but perhaps right now [there's no other way](https://github.com/NVIDIA/cuQuantum/discussions/85). I can chuck this in during my GPU patches, which I expect is a one-liner like:
  ```cmake
  target_include_directories(QuEST PRIVATE $ENV{CUQUANTUM_ROOT})
  ```


Thanks very much for the fantastic work!! Will merge right away :tada:


*although I have to correct a validation error message in deprecated `destroyComplexMatrixN`, grr

otbrown:
Woooooo, fantastic news! Thanks for all your reviews, and of course for developing QuEST v4 ;)

Your one-liner should indeed do the trick, although you may want that include to be `PUBLIC` so executables built against it can also find the header. 

The more modern CMake way is for me to grumpily write a `FindCuQuantum.cmake` file on NVIDIA's behalf and create a target. Shouldn't take too long, just need to install cuquantum somewhere to test it.

TysonRayJones:
@otbrown Did you encounter any issues with Catch2 versions more recent than `v3.1.1` in the current build? I really want to use the [`SKIP`](https://github.com/catchorg/Catch2/blob/devel/docs/skipping-passing-failing.md) macro introduced in `v3.3.0` which seems to be downloading/compiling fine on MacOS and Ubuntu, and working with both the deprecated v3 tests and the v4 stubs. Compiling v3.3.0 is noticeably slower than v3.1.1 alas, but that's no biggie.

I'm keen to use `SKIP` because the tests will now be performed on at least _eight_ unique Qureg deployment combinations, most of which will be skipped when run by users who do not have a multi-GPU multi-core machine hehe.

TysonRayJones:
No need to make the change by the way - in fact, I might not end up using `SKIP` (because I'm unhappy with the facilities for suppressing polluting "skipped" messages). But still interested if there was an obstacle!

otbrown:
Nope, none at all! In fact my local version was 3.4 and it worked fine.

Feel free to use whatever newer functionality you like, and then we can bump the minimum version accordingly :grin:

otbrown:
Compiler support is pretty green across [the board](https://en.cppreference.com/w/cpp/compiler_support#cpp20) these days for C++20 these days, so I don't think there's a good argument against this except it makes me feel old 😆

otbrown:
Watch out for the "partial" compliance in clang though!

TysonRayJones:
wew nice, ty ty! I'll work locally with clang 🙏

PR #436, opened 2024-07-10
added complex arithmetic

TysonRayJones:
in a precision agnostic fashion. This enables C and C++ users to perform arithmetic between qcomp and non-equal-precision numbers (like integers, non-qreal decimals, etc)

Other changes:
- added reportQcomp
- added arithmetic examples
- cleaned up formatter.cpp, overloading form_str to format complex numbers so that other core files can stringify complex numbers

### DIFF
--- a/examples/numbers/arithmetic.c
+++ b/examples/numbers/arithmetic.c
@@ -2,7 +2,8 @@
 
 int main() {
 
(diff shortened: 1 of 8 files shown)


TysonRayJones:
Very much don't love that refreshing the page (to load the final commit) changed my chosen "squash" option to "rebase" grr

PR #444, opened 2024-07-25
Pauli data structures

TysonRayJones:


TysonRayJones:
just need new example files

PR #452, opened 2024-08-06
testing generic indexing

TysonRayJones:
Macbook, single threaded:

5 qb:
dur_OLD_noCtrlGate:	12355
dur_NEW_noCtrlGate:	19522

dur_OLD_oneCtrlGate:	37503
dur_NEW_oneCtrlGate:	36666

dur_OLD_oneCtrlGate:	66097
dur_NEW_oneCtrlGate:	55265

7 qubit:
dur_OLD_noCtrlGate:	52402
dur_NEW_noCtrlGate:	71317

dur_OLD_oneCtrlGate:	85830
dur_NEW_oneCtrlGate:	89545

dur_OLD_oneCtrlGate:	155343
dur_NEW_oneCtrlGate:	131910

10 qubit
dur_OLD_noCtrlGate:	788733
dur_NEW_noCtrlGate:	792376

dur_OLD_oneCtrlGate:	537364
dur_NEW_oneCtrlGate:	531666

dur_OLD_oneCtrlGate:	695470
dur_NEW_oneCtrlGate:	627380

The slowdowns could be explained by:

- NEW inserted target bit it always a runtime variable, whereas OLD was a compile-time literal (at least for the no-ctrl case). We can easily remedy this by having the getNthInd() functions have a few hand-optimised base-cases using literals. BUT it'd be nice if we could somehow achieve the same thing using const, or passing arr[NumQubits] instead of vectors to getNthInd() functions, etc

- struct fixed-size array access (even when only 2 elements long) is still slower than variable access? Surely not! Surely the compiler is smart enough to see the whole array can fit into registers

- the automatic loop unrolling has some pitfall? Would using OpenMP's `#pragma omp unroll full` improve it when multithreaded?


TysonRayJones:
**Victor, serial**

5 qubits

```
dur_OLD_noCtrlGate:	29277
dur_NEW_noCtrlGate:	30222

dur_OLD_oneCtrlGate:	28196
dur_NEW_oneCtrlGate:	28458

dur_OLD_oneCtrlGate:	35995
dur_NEW_oneCtrlGate:	34386
```

10 qubits:
```
dur_OLD_noCtrlGate:	1736666
dur_NEW_noCtrlGate:	1760713

dur_OLD_oneCtrlGate:	954470
dur_NEW_oneCtrlGate:	958885

dur_OLD_oneCtrlGate:	990238
dur_NEW_oneCtrlGate:	987752
```

Woohoo! This already assures us that for few qubits, the use of a struct-array and relying on automatically-unrolled for loop is equivalent to hand-optimised variables.

**Victor, 12 threads**

10 qubits:
dur_OLD_noCtrlGate:	422616
dur_NEW_noCtrlGate:	432849

dur_OLD_oneCtrlGate:	381984
dur_NEW_oneCtrlGate:	373267

dur_OLD_oneCtrlGate:	433478
dur_NEW_oneCtrlGate:	424242

TysonRayJones:
It looks instead as if Clang is able to make additional, wildly effective optimisations when the inserted bit is a literal (single target, no control), making my 2017 Macbook twice as fast as Victor! Super interesting!

PR #454, opened 2024-08-18
changed communication buffer indexing

TysonRayJones:
so that the sent sub-buffers always begin in the middle of the full-buffer. This enables the communicated amplitudes to be received to the first index (zero), to be consistent with the scenario when sub-buffers are not packed (the buffer is filled straight from the Qureg amplitudes). As a result, post-communication subroutines always know where the received amplitudes begin in the buffer; from zero.

This is crucial; many post-communication routines are agnostic to the number of control qubits, and as a result, they do not know whether all amplitudes were exchanged (the buffer was filled entirely) or a sub-buffer was packed and exchanged. Before this change, the receiving index was inconsistent between these two situations, requiring the post-comm subroutine to condition on the number of suffix controls, and was liable to a bug

### DIFF
--- a/quest/src/cpu/cpu_subroutines.cpp
+++ b/quest/src/cpu/cpu_subroutines.cpp
@@ -16,7 +16,7 @@
 #include "quest/src/core/bitwise.hpp"
 #include "quest/src/core/utilities.hpp"
 #include "quest/src/core/accelerator.hpp"
-#include "quest/src/comm/comm_routines.hpp"
+#include "quest/src/comm/comm_indices.hpp"
 
 #include <vector>
 
@@ -107,7 +107,8 @@ void cpu_statevec_anyCtrlOneTargDenseMatr_subB(Qureg qureg, vector<int> ctrls, v
     // each control qubit halves the needed iterations
     qindex numIts = qureg.numAmpsPerNode / powerOf2(ctrls.size());
     
-    qindex offset = getSubBufferRecvInd();
+    // received amplitudes may begin at an arbitrary offset in the buffer
+    qindex offset = getBufferRecvInd();
 
     auto sortedCtrls   = util_getSorted(ctrls);
     auto ctrlStateMask = util_getBitMask(ctrls, ctrlStates);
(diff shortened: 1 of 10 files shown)


quest/src/cpu/cpu_subroutines.cpp
```diff
@@ -36,6 +37,9 @@ void cpu_statevec_packAmpsIntoBuffer(Qureg qureg, vector<int> ctrls, vector<int>
     // each control qubit halves the needed iterations
     qindex numIts = qureg.numAmpsPerNode / powerOf2(ctrls.size());
 
+    // amplitudes are packed at an offset into the buffer
+    qindex offset = getSubBufferSendInd(qureg);
```
TysonRayJones, line 41: unsure whether knowing `offset=0` at compile-time offers any measurable speedup

quest/src/cpu/cpu_subroutines.cpp
```diff
@@ -115,11 +121,7 @@ void cpu_statevec_anyCtrlOneTargDenseMatr_subB(Qureg qureg, vector<int> ctrls, v
         // i = nth local index where ctrl bits are in specified states
         qindex i = insertBitsWithMaskedValues(n, sortedCtrls.data(), numCtrlBits, ctrlStateMask);
 
-        // l = index of nth received buffer amp
-        qindex l = n + numIts;
-        qcomp amp = qureg.cpuCommBuffer[l];
-
-        qureg.cpuAmps[i] = fac0*qureg.cpuAmps[i] + fac1*amp;
+        qureg.cpuAmps[i] = fac0*qureg.cpuAmps[i] + fac1*qureg.cpuCommBuffer[offset+n];
```
TysonRayJones, line 124: unsure whether knowing `offset=0` at compile-time offers any measurable speedup

TysonRayJones:
Before merging this, I wish to test whether the pedantic and defensively-designed runtime variable [`offset`](https://github.com/QuEST-Kit/QuEST/pull/454/files#r1721049167) in the CPU hot-loop has any performance effect; my guess is _absolutely not_.

If it _does_ degrade performance, and we wish to maintain the defensive-design, we will move the definition of `getSubBufferRecvInd()` into a new `comm_indexer.hpp` header and declare it as `constexpr` or `INLINE` (via `inliner.hpp`)

PR #532, opened 2025-01-31
cuQuantum in CMake

otbrown:
Me again!

This integrates cuQuantum into the CMake in a slightly hacky way, but at least enables the dependency to be propagated to binaries using QuEST. I've checked it works okay on Cirrus, but let me know if you encounter issues.

Users should be able to either export `CUQUANTUM_ROOT` in their environment, or use the more CMake way and define `CUQUANTUM_DIR` at configure time.


TysonRayJones:
Woo brilliant! I've confirmed this works on Ubuntu, and that the error message is triggered when I mutilate `CUQUANTUM_ROOT` 🎉

PR #539, opened 2025-02-18
Fixing MSVC Compiler Errors

### LINKED ISSUE #538: MSVC fails to build QuEST
Hi
I've been trying to build the v4 branch on Windows using MSVC, but I've encountered compiler issues. Below is an example of the errors I initially received:

```shell
E:\projects\QuEST\quest\include\matrices.h(220,9): error C7555: use of designated initializers requires at least '/std:c++20' [E:\projects\QuEST\cmake-build-dev\QuEST.vcxproj]
  (compiling source file '../quest/src/core/memory.cpp')
```

At first, I thought that upgrading the C++ standard to C++20 in CMake would solve the issue, but that wasn't the case. The compiler errors then changed to:

```shell
E:\projects\QuEST\quest\include\matrices.h(219,12): error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax [E:\projects\QuEST\cmake-build-dev\QuEST.vcxproj]
  (compiling source file '../quest/src/api/calculations.cpp')
```

Digging a bit deeper, it appears that the following class of functions is using non-standard syntax that MSVC rejects:

```c++
static inline CompMatr1 getCompMatr1(qcomp** in) {

    return (CompMatr1) {
        .numQubits = 1,
        .numRows = 2,
        .elems = {
            {in[0][0], in[0][1]}, 
            {in[1][0], in[1][1]}}
    };
}
```
Designated initializers are a C++20 feature, but combining them with a C-style cast seems to result in undefined behaviour. I believe that removing the casts from these functions and bumping the standard to C++20 will fix the problems with MSVC.
Cheers,
Erich

eessmann:
**Fixes #538**

### Summary

This PR resolves compiler errors on MSVC by replacing non-standard usage of designated initializers combined with C-style casts. Previously, the code used:

```c++
return (CompMatr1) {
    .numQubits = 1,
    .numRows = 2,
    .elems = {
        {in[0][0], in[0][1]},
        {in[1][0], in[1][1]}
    }
};
```

which is incompatible with MSVC. The updated implementation ensures standard compliance by using:

```c++
CompMatr1 out = {
    .numQubits = 1,
    .numRows = 2,
    .elems = {
        {in[0][0], in[0][1]},
        {in[1][0], in[1][1]}
    }
};
return out;
```

Additionally, functions relying on GCC-specific intrinsics have been refactored to improve cross-compiler compatibility.

### Background

Attempting to build the v4 branch on Windows with MSVC resulted in the following errors:

```
E:\projects\QuEST\quest\include\matrices.h(220,9): error C7555: use of designated initializers requires at least '/std:c++20'
```

and

```
E:\projects\QuEST\quest\include\matrices.h(219,12): error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax
```

These errors occurred due to a combination of designated initializers and a C-style cast, which MSVC does not support, even with C++20 enabled. Removing the cast resolves this issue.

### Testing & Known Issues

- **Static Library Mode:** Successfully builds and passes all tests.
- **Shared Library Mode:** The build fails and requires further investigation.

Cheers,  
Erich

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -154,6 +154,12 @@ option(
 )
 message(STATUS "Deprecated API support is turned ${ENABLE_DEPRECATED_API}. Set ENABLE_DEPRECATED_API to modify.")
 
+# Windows Specific Options
+if(MSVC)
+    set(ENABLE_MULTITHREADING OFF)
+    set(BUILD_SHARED_LIBS OFF)
+endif()
+
 ## Library
 
 add_library(QuEST)
@@ -175,7 +181,7 @@ target_include_directories(QuEST
 target_compile_features(QuEST
   PUBLIC
   c_std_11
-  cxx_std_17
+  cxx_std_20
 )
 
 # Turn on all compiler warnings
@@ -286,7 +292,7 @@ add_subdirectory(quest)
 add_executable(min_example
   examples/tutorials/min_example.cpp
 )
-target_link_libraries(min_example PUBLIC QuEST::QuEST)
+target_link_libraries(min_example PRIVATE QuEST::QuEST)
 
 install(TARGETS min_example
   RUNTIME
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,6 +157,7 @@ message(STATUS "Deprecated API support is turned ${ENABLE_DEPRECATED_API}. Set E
 # Windows Specific Options
 if(WIN32)
     set(ENABLE_MULTITHREADING OFF)
+    message(WARNING "multithreading disabled on Windows")
     # Force MSVC to export all symbols in a shared library, like GCC and clang
     set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
     if (ENABLE_TESTING)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,7 @@ set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake" ${CMAKE_MODULE_PATH})
 include(GNUInstallDirs)
 
 # Maths
-if (NOT MSVC)
+if (NOT WIN32)
   find_library(MATH_LIBRARY m REQUIRED)
 endif()
 
@@ -155,9 +155,13 @@ option(
 message(STATUS "Deprecated API support is turned ${ENABLE_DEPRECATED_API}. Set ENABLE_DEPRECATED_API to modify.")
 
 # Windows Specific Options
-if(MSVC)
+if(WIN32)
     set(ENABLE_MULTITHREADING OFF)
-    set(BUILD_SHARED_LIBS OFF)
+    # Force MSVC to export all symbols in a shared library, like GCC and clang
+    set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
+    if (ENABLE_TESTING)
+      set(BUILD_SHARED_LIBS OFF)
+    endif ()
 endif()
 
 ## Library
--- a/quest/include/matrices.h
+++ b/quest/include/matrices.h
@@ -216,18 +216,19 @@ typedef struct {
 
 static inline CompMatr1 getCompMatr1(qcomp** in) {
 
-    return (CompMatr1) {
+     CompMatr1 out = {
         .numQubits = 1,
         .numRows = 2,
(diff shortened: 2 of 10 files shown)


TysonRayJones:
Brilliant, excellent stuff!! Good catch on de-mangling `invalidQuESTInputError` too. I've confirmed this is all compiling fine on MacOS and Ubuntu, in C++20 and C11. 

A shame to have to disable multithreading altogether on Windows. Should we try to address it? It results from _eight_ OpenMP reductions in [cpu_subroutines](https://github.com/QuEST-Kit/QuEST/blob/25256814bfe143866541d1afd971b55032b68919/quest/src/cpu/cpu_subroutines.cpp#L1768) making use of the [`qcomp` custom reduction](https://github.com/QuEST-Kit/QuEST/blob/25256814bfe143866541d1afd971b55032b68919/quest/src/cpu/cpu_subroutines.cpp#L34). I was so far _ideologically_ recoiling from the simple workaround of reducing each component:
```C++
    qreal prodRe = 0;
    qreal prodIm = 0;

    #pragma omp parallel for reduction(+:prodRe,prodIm)
    for (qindex n=0; n<numIts; n++) {
        qcomp term += conj(quregA.cpuAmps[n]) * quregB.cpuAmps[n];

        // reduce components separately because MSVC OpenMP is stanky
        prodRe += real(term);
        prodIm += imag(term);
    }

    return qcomp(prodRe,prodIm);
```
I could be persuaded to this evil _if_ it has no performance effect on the other compilers. Could it possibly interfere with auto-vectorisation or something?

(ignore the irrelevant/outdated v3 workflows, I was just curious as to why I was prompted to run them).

eessmann:
Hi Tyson,

I’ve tracked down and fixed the root cause of the shared library build failures on Windows. By default, Windows doesn’t export any symbols in DLLs, unlike GCC/Clang on Linux. Fortunately, CMake provides a simple option to force symbol exports:

```cmake
# Windows-specific options
if(WIN32)
    set(ENABLE_MULTITHREADING OFF)
    # Force MSVC to export all symbols when building a shared library
    set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
    if (ENABLE_TESTING)
        # Catch2 has issues being built as a shared library on Windows,
        # so use static libraries when building tests
        set(BUILD_SHARED_LIBS OFF)
    endif()
endif()
```

This resolves the shared library build issues. However, because Catch2 doesn’t play nicely as a DLL on Windows, I switched to static libraries for the testing build as a temporary workaround.

Regarding OpenMP, I spent some time trying to get it working under MSVC (including trying the new [LLVM-based backend](https://learn.microsoft.com/en-us/cpp/build/reference/openmp-enable-openmp-2-0-support?view=msvc-170)). Unfortunately, none of these attempts solved our existing compatibility problems. There doesn’t seem to be a quick or cost-free solution here, so we may need to accept that OpenMP support on Windows remains disabled for the time being.

Thanks again,  
Erich

TysonRayJones:
Sure, making Catch2 static on Windows is inoffensive enough.

Ahh I see. Are there issues with OpenMP on Windows _beyond_ the custom reduction? If so, then I agree with disabling for the moment. In any case, we should add a warning during the CMake build to alert multithreading has been forcefully disabled, so Windows users aren't astonished by poor performance.

TysonRayJones:
Hi Erich,
Oops I clicked 'ready for review' in curiosity of what would happen given I didn't author the PR hehe. Is this ready to merge into v4? If so, I'll test it on my systems and add a "multithreading disabled on Windows"-esque alert to the build

eessmann:
Hi Tyson,
I think it's ready to merge. Its a pitty about the lack of OpenMP on windows. I checked and MPI seems to work on MSVC so we have a options for parallel processing on windows.

TysonRayJones:
Great, all merged! Alas MPI isn't a sufficient replacement for OpenMP on Windows, since some downstream projects (like [QuESTlink](https://github.com/QTechTheory/QuESTlink), for which there is a surprisingly large Windows userbase) cannot work with MPI. 

I'll try restoring MSVC multithreading in QuEST v4.1 - it may just need some ugly changes (e.g. avoiding `qcomp` reduction) on par with similarly ugly changes needed to restore AMD compatibility being considered [here](https://github.com/otbrown/QuEST/pull/2).

PR #535, opened 2025-02-09
v4 unit tests

TysonRayJones:

PR #541, opened 2025-02-19
more patches

TysonRayJones:
awaiting anticipated cuQuantum patches (testing on DiRAC)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -175,7 +175,7 @@ target_include_directories(QuEST
 target_compile_features(QuEST
   PUBLIC
   c_std_11
-  cxx_std_17
+  cxx_std_20
 )
 
 # Turn on all compiler warnings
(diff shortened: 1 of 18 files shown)


TysonRayJones:
do not merge this until overlapping commits in #535 have been reversed

TysonRayJones:
As of now, all new unit tests (which check only operations, e.g. `apply*`) are passing on all backends (i.e. including cuQuantum). However, the deprecated tests have not been tested with cuQuantum, and the deprecated tests also exclude some critical functions for cuQuantum (like `calcExpecPauliStr`)

TysonRayJones:
Failing cuQuantum deprecated tests:
- measureWithStats
- measure
- collapseToOutcome
(all such tests pass with non-cuQuantum GPU backend)

TysonRayJones:
Note non-cuQuantum GPU + 8-node MPI encountered seg-fault during deprecated `calcPurity` on DiRAC (A100 CC=80), but which is not encountered on Victor (Quadro P6000 CC=61). It's also not encountered in similarly distributed CPU mode.

Enjoy debugging that, grrr!

TysonRayJones:
There's a remaining bug in cuQuantum probability calculations (like `calcProbOfQubitOutcome`) causing a crash with "not initialized" error. We defer the patch so that v4 can compile and run some basic tests now, for ease of development

PR #550, opened 2025-03-03
added code coverage CI

TysonRayJones:


TysonRayJones:
code coverage too slow - move to paid long runner

PR #554, opened 2025-03-05
CMake: minor tweak to hip compilation

otbrown:
Hi Tyson, 

Without wishing to traumatise you further, I thought I'd have a look into the AMD GPU CI failures, you were battling yesterday. I can't see any obvious fix unfortunately. One thing to try might be adding:
```
find_package(rocthrust REQUIRED)
target_link_libraries(QuEST PRIVATE roc::rocthrust)
```
however, when I tried that on A2 it a) broke the ability for us to compile the non-GPU code with anything but hipcc, as it adds the offload target flag everywhere, and b) lead to an unexpected link-time failure. 

It might be more interesting to just try `find_package(rocthrust REQUIRED)` to check that the build system does actually see that package _somewhere_. From brief inspection it looks like rocThrust is a header only template libary, so there shouldn't need to be any actual linkage, just the right include directories. I just ran compilation with `-H` and we for sure are picking up these includes from the generic hip target on ARCHER2, so I would _expect_ the same to be true elsewhere, but this could be ROCm version dependent, which may be the issue on the CI. I can see in the logs that `rocmthrust-dev` is on the install list so it would be surprising.

Anyway -- the two actual things which are in this PR!
- ROCm documentation suggests using `hip::host` as the link target. I checked the CMake files and it is just an interface target to `hip::amdhip64`, so it should (and does as far as I can tell) generate the same build files, but ought to be more robust across platforms as it's what AMD expects. 
- I removed the definition of `__HIP_PLATFORM_AMD__` as CMake adds this anyway when the target language is HIP, so it ends up double defined in the compile line currently. That doesn't break anything, but it's just cruft.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -182,7 +182,7 @@ jobs:
           -DCMAKE_CUDA_ARCHITECTURES=${{ env.cuda_arch }}
           -DCMAKE_HIP_ARCHITECTURES=${{ env.hip_arch }}
           -DCMAKE_CXX_COMPILER=${{ matrix.compiler }}
-          -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=${{ matrix.mpi == 'ON' && matrix.cuda == 'ON' && 'OFF' || 'ON' }}
+          -DINTERPROCEDURAL_OPTIMIZATION=${{ matrix.mpi == 'ON' && matrix.cuda == 'ON' && 'OFF' || 'ON' }}
 
       # force 'Release' build (needed by MSVC to enable optimisations)
       - name: Compile
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -159,6 +159,22 @@ jobs:
           sudo apt install amdgpu-dkms rocm
           echo "${{ env.rocm_path }}" >> $GITHUB_PATH
 
+      # adjust compiler flags (in OS-agnostic fashion)
+      - name: Adjusting compiler and flags
+        run: |
+          echo "flags=''" >> $GITHUB_ENV
+          echo "compiler=${{ matrix.compiler }}" >> $GITHUB_ENV
+      - name: Disabling LTO on MPI + CUDA
+        if: ${{ matrix.mpi == 'ON' && matrix.cuda == 'ON' }}
+        run: |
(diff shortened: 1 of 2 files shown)


TysonRayJones:
I had the same experience on A2 trying to link `hiprtc`; `hipcc` compiled all src fine (after updating `cpu_config.cpp`) but failed at link time with indecipherable vtable errors. But _maybe_ that's a symptom of some A2 issue that wouldn't occur in a github VM 🥲 

I will try two things on this branch, restoring HIP + MPI CI attempts:
- including `find_package(rocthrust REQUIRED)`
- above, but additionally using `hipcc` to compile src like some of my miserable, deranged, embarrassing [attempts](8b0543327b4614142789e94a3f01d93f45e05117) yesterday

If these mere two attempts fail, I will _immediately_ give up and revert.

otbrown:
Sounds good! Definitely not worth spending more time on 😆 At least it seems to work on ARCHER2. 
The other thing from the logs that's somewhat suspect is:
```
CMake Warning (dev) at /opt/rocm/lib/cmake/hip/hip-config-amd.cmake:91 (message):
   AMDGPU_TARGETS was not set, and system GPU detection was unsuccsesful.
   
   The amdgpu-arch tool failed:
   Error: 'Failed to get device count'
   Output: ''
   
   As a result, --offload-arch will not be set for subsuqent
   compilations, and the default architecture
   (gfx906 for dynamic build / gfx942 for static build) will be used
   ```
   
 I can sort of imagine a situation in which somehow not providing the right arch breaks the includes. It is absolutely not clear to me why CMAKE_HIP_ARCHITECTURES is being ignored however. Possibly an interaction with CMAKE_CUDA_ARCHITECTURES as I think technically HIP is "platform agnostic"? In any case, it might be worth setting `GPU_TARGETS` from the ROCm documentation, or `AMDGPU_TARGETS` recommended in the error message.

otbrown:
Oh final bit of CMake magic for now -- [CMAKE_INERPROCEDURAL_OPTIMIZATION](https://cmake.org/cmake/help/latest/variable/CMAKE_INTERPROCEDURAL_OPTIMIZATION.html#cmake-interprocedural-optimization)=False can be used to turn off LTO, which I see has come up a few times in the GitHub runners 😅

PR #558, opened 2025-03-07
more GPU bug patches

TysonRayJones:
can ya bilivit


quest/src/api/environment.cpp
otbrown, line 1: Ehehe, was just coming over to report the GPU runtime failure when a GPU is present but not compiled, and you've already fixed it 🧙

TysonRayJones: hehe sorry about that! I'm embarrassed to have committed tests to the `v4` branch which fail in certain hardware configs, but there've been too many bugs to juggle. 

As of now, every _existing_ v4 test (noting only about half the API is tested) is passing in every configuration, _except_ NVLlinked; I haven't revisited a troubling error I saw a while ago. Might even be solved now though 🙏  Going to complete the important v4 tests before I return to it

PR #563, opened 2025-03-11
scaffolded v4 doc

TysonRayJones:


TysonRayJones:
This PR could not be squash-merged (despite that I tried several times) so was manually squashed and rebased - which github forced into an ugly merge-commit

PR #584, opened 2025-04-05
added C++ convenience overloads

TysonRayJones:


TysonRayJones:
We do these things not because they are easy, but because we thought they were going to be easy

Issue #592, opened 2025-04-22
Issues with running compiled examples on Windows


TysonRayJones:
Compiling on Windows (with MSVC) via (from `build\`)
```
cmake .. -D BUILD_EXAMPLES=ON -D CMAKE_BUILD_TYPE=RELEASE
cmake --build . --config Release
```
has been reported to compile to
```
build\Release\min_example.exe
```
which can be run. But alas the other examples, such as
```
build\examples\matrices\Release\cpp_initialisation.exe
```
fail to execute because `"QuEST.dll cannot be found"`.

This requires:
- [x] correcting `compile.md` with the location of executables on Windows (the additional `\Release\` subdirectory)
- [x] correcting `launch.md` similarly
- [x] resolving the missing DLL

Note example compilation now differs on the `devel` branch from when this bug was found but likely persists.

> Found by Joshua Dai :tada:

TysonRayJones:
- [x] Add also mention of [`CMAKE_RUNTIME_OUTPUT_DIRECTORY`](https://cmake.org/cmake/help/latest/variable/CMAKE_RUNTIME_OUTPUT_DIRECTORY.html) to compile and launch doc (_not_ Windows specific)

TysonRayJones:
Patched by Erich in 1c051cf

PR #601, opened 2025-04-30
RPATH handling and installation paths fix

eessmann:
Hi Tyson
This code adjusts the install interface paths to use CMAKE_INSTALL_INCLUDEDIR and adds RPATH handling using the helper function setup_quest_rpath. This ensures consistent RPATH behaviour across platforms.
Erich


TysonRayJones:
Wew thanks very much for looking at this! Could you PR this instead to the [`devel`](https://github.com/QuEST-Kit/QuEST/tree/devel) branch? Then we can easily test it with changes made to the examples building, and bundle it with a `v4.1` merge into `main`.

eessmann:
Sure, no problem
The RPATH fix should clear up some issues I have been having on my quest rust bindings side project

TysonRayJones:
Thanks very much! Btw do you think it's worth testing _execution_ of the executables compiled during CI? The compiled examples (at least, those in `examples/isolated`) run in mere seconds, but could reveal a build issue like the bundling of the DLLs (as per #592).

If that's worthwhile, we could bundle the workflow change into this PR to test these build changes "for free". It _should_ be as simple as adding code like below to the bottom of [`compile.yml`](https://github.com/QuEST-Kit/QuEST/blob/devel/.github/workflows/compile.yml):
```yml
      # attempt to run the compiled executable, testing e.g. DLL location
      - name: Run
        if: ${{ matrix.os == 'windows-latest' }}
        shell: bash
        run: |
          ./${{ env.build_dir }}/examples/isolated/Release/complex_arithmetic_c.exe
      - name: Run
        if: ${{ matrix.os != 'windows-latest' }}
        run: |
          ./${{ env.build_dir }}/examples/isolated/complex_arithmetic_c
```

If you think it's a good idea, I'll PR to your `rpath-fix` branch and we can squash-commit this PR together. Alternatively, you can make the above above edits yourself to [`compile.yml`](https://github.com/QuEST-Kit/QuEST/blob/devel/.github/workflows/compile.yml), though that might prove a headache; it will require my manually re-triggering the CI (auto-permissions disabled until the paid runners aren't on my credit card hehe), and experiencing the general horror of tinkering with Github Actions.

eessmann:
I think it's a great idea to run an isolated exe for a few secs in CI. It should catch most link time errors

PR #614, opened 2025-05-06
V4 header install

### LINKED ISSUE #611: Installing headers
I have a few issues with installing QuEST v4 header files.
1. Headers are installed to the configured destdir, not the one passed at install-time, which means the build directory needs to be re-configured at every different install path, and that if you did not specify the header install dir at all initially it will attempt to install at the system location and fail:
  ```
cmake --install build/ --prefix $HOME/quantsim/versions/v4-main
-- Install configuration: "RelWithDbgInfo"
-- Installing: /usr/local/quest/include
CMake Error at build/quest/include/cmake_install.cmake:54 (file):
  file INSTALL cannot make directory "/usr/local/quest/include": No such file
  or directory.
Call Stack (most recent call first):
  build/quest/cmake_install.cmake:52 (include)
  build/cmake_install.cmake:47 (include)
  ```

2. Headers aren’t installed under `$PREFIX/include` but under `$PREFIX/quest/include` because that seems to be their path inside the source distribution
3. Once installed, including headers fails as they use hard-coded `quest/include/` prefixes. This means you need both a `-I$PREFIX/quest/include` to find `quest.h` and then a `-I$PREFIX` so that `quest.h`'s includes can find the other quest includes.
4. Once installed, `quest/include/modes.h` are still requiring `COMPILE_XXX`  to be set, but those are compile-time constants. The values set at compile-time should be exported in a header file and included along with the rest so that they are coherent with the installed binaries. I know that currently they are exported in the `INTERFACE_COMPILE_DEFINITIONS` in `QuESTTargets.cmake` but that restricts any usage to CMake projects.

I think it would be sufficient to adjust the install paths for quest headers to be under `$PREFIX/include` (keeping relative directories so `$PREFIX/include/quest/include/*.h`) and adding a `quest.h.in` that goes to `$PREFIX/include/quest.h`, that receives the configuration values and includes the previous `quest.h`. Or otherwise a more standard `config.h.in` and correcting the layout of includes.

lucjaulmes:
Fixes #611 header issues:
- Moves the install directive for headers into main CMakeLists to make it recognize the install prefixes correctly
- Create  a more standard layout with `quest.h` and other headers under `quest/` subdirectory all in the usual include path.
- Use the build directories to create a structure similar to the desired install directories so that include paths are unique.
- Make `quest.h` configured so it contains the pre-processor defines if `VERBOSE_LIB_NAME` is false, and make those target defines private so they don’t appear in exported cmake files.
- Do not set the defines in `quest.h` and keep them in the exported cmake if `VERBOSE_LIB_NAME` is true.
- Differentiate exported cmake files using `$LIB_NAME` as well, so the matching cmake files link to the right verbosely-named libraries and provide the right compile-time flags, with a single set of header files.

I think all of the CMakeFiles machinery (in particular `foreach()` which seems to be discouraged) can be simplified at the cost of having all the files already in the final layout in the source directory. So all headers at `quest/include/quest.h.in` and `quest/include/quest/*.h` respectively. I can update the PR that way if that’s preferred.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -525,7 +525,7 @@ set(QuEST_INSTALL_CONFIGDIR "${CMAKE_INSTALL_LIBDIR}/cmake/QuEST")
 
 # Write QuESTConfigVersion.cmake
 write_basic_package_version_file(
-        "${CMAKE_CURRENT_BINARY_DIR}/QuESTConfigVersion.cmake"
+        "${CMAKE_CURRENT_BINARY_DIR}/${LIB_NAME}ConfigVersion.cmake"
         VERSION ${PROJECT_VERSION}
         COMPATIBILITY AnyNewerVersion
 )
@@ -533,14 +533,14 @@ write_basic_package_version_file(
 # Configure QuESTConfig.cmake (from template)
 configure_package_config_file(
         "${CMAKE_CURRENT_SOURCE_DIR}/cmake/QuESTConfig.cmake.in"
-        "${CMAKE_CURRENT_BINARY_DIR}/QuESTConfig.cmake"
+        "${CMAKE_CURRENT_BINARY_DIR}/${LIB_NAME}Config.cmake"
         INSTALL_DESTINATION "${QuEST_INSTALL_CONFIGDIR}"
 )
 
 # Install them
 install(FILES
-        "${CMAKE_CURRENT_BINARY_DIR}/QuESTConfig.cmake"
-        "${CMAKE_CURRENT_BINARY_DIR}/QuESTConfigVersion.cmake"
+        "${CMAKE_CURRENT_BINARY_DIR}/${LIB_NAME}Config.cmake"
+        "${CMAKE_CURRENT_BINARY_DIR}/${LIB_NAME}ConfigVersion.cmake"
         DESTINATION "${QuEST_INSTALL_CONFIGDIR}"
 )
 
@@ -554,7 +554,7 @@ install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/quest/include/"
 )
 
 install(EXPORT QuESTTargets
-        FILE QuESTTargets.cmake
+        FILE "${LIB_NAME}Targets.cmake"
         NAMESPACE QuEST::
         DESTINATION "${QuEST_INSTALL_CONFIGDIR}"
 )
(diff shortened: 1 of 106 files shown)


TysonRayJones:
Hi Luc,

Thanks very much for the PR! I like your approach, though am still investigating the problem. 

If possible, I'd wish to avoid changing the hundreds of include paths. I've so far been using a variant of [Kolpackov's structure](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html), using the include path to semantically separate what's frontend and backend. Consider the includes in [`accelerator.cpp`](https://github.com/QuEST-Kit/QuEST/blob/main/quest/src/core/accelerator.cpp):
```C++
#include "quest/include/types.h"
#include "quest/include/qureg.h"
#include "quest/include/paulis.h"
#include "quest/include/matrices.h"

#include "quest/src/core/accelerator.hpp"
#include "quest/src/core/errors.hpp"
#include "quest/src/core/memory.hpp"
#include "quest/src/core/bitwise.hpp"
#include "quest/src/cpu/cpu_config.hpp"
#include "quest/src/gpu/gpu_config.hpp"
#include "quest/src/cpu/cpu_subroutines.hpp"
#include "quest/src/gpu/gpu_subroutines.hpp"

#include <vector>
#include <algorithm>
```

A change like this PR's then seems to mislead about the location of the headers
```C++
#include "quest/types.h"
...

#include "quest/src/core/accelerator.hpp"
...
```
It also creates an asymmetry in the working directory of frontend and backend includes within a backend file.

No need to explore this further yourself, I'll tinker when I get a chance to see if the existing paths can be preserved without being terribly non-standard and hacky. 

Finally, note the PR should point to `devel` rather than `main` 🙏

lucjaulmes:
Sure, no worries! I’ll close this down as it’s pointing to the wrong branch. Just thought I’d share because this is what I’m using currently to run v4 and install different versions to different prefixes on the same machine.

It seems to me though that the structure you’re referring to recommends includes of the type `quest/types.h` as well as stripping `quest/src/` from the private includes (and having `$CMAKE_SOURCE_DIR/quest/src/` added to CMake include paths instead of the `$CMAKE_SOURCE_DIR`). But that’s even more include path changing.

You should be able install to headers as `$prefix/include/quest/include/types.h` (just a bit odd and non-standard paths), and then create a `quest.h.in` that works from `$CMAKE_BINARY_DIR/include` for minimal changes in all layout and includes. In fact I don’t think it would take too much time for me to adjust this PR that way if you’d want to see it.

otbrown:
I see Tyson has beat me to it, but chiming in to say thanks for this @lucjaulmes! Especially for solving the headache of leaking compile definitions, which I'd been sitting on until I had more time.

Apologies for not picking up the install bug sooner, but as ever it worked on my ~~machine~~ workflow! 😅 

RE: foreach, I think that and a few other issues could be solved using file sets, except we would need to bump to CMake 3.23. I can look at that later, no need to do it all once in this PR.

---

Anddddd I was still too slow 😆 @lucjaulmes yes, if you have time to make the revised PR to `devel` that would be appreciated, but no worries if not!

PR #630, opened 2025-05-28
Optimise calcExpecPauliStrSum() by batch-processing similar PauliStr …

### LINKED ISSUE #599: Optimise `calcExpecPauliStrSum()` by batch-processing similar `PauliStr`
## Summary

Optimise the statevector backend(s) of [`calcExpecPauliStrSum()`](https://github.com/QuEST-Kit/QuEST/blob/53f8f3ad60e5b0171646ee8250c0e6ea65878b44/quest/src/api/calculations.cpp#L146) by grouping constituent Pauli strings by the amplitudes they combine.

## Context

Hermitian observables are canonically expressed as real-weighted sums of Pauli strings:

$$
\hat{H} = \sum\limits_i^T c_i \\, \hat{\sigma}_i, \\,\\,\\,\\,\\,\\, c_i \in \mathbb{R}, \\,\\,\\,\\,\\,\\, \hat{\sigma}_i \in \\{\hat{I},\hat{X},\hat{Y},\hat{Z}\\}^{\otimes N}.
$$

For example, consider

$$
\hat{H} = 
1.5 \\, (\hat{X} \otimes \hat{Y} \otimes \hat{Z}) + 
0.3 \\, (\hat{Z} \otimes \hat{Z} \otimes \hat{X}) - 
0.2 \\, (\hat{Z} \otimes \hat{Z} \otimes \hat{Y}).
$$

We often seek the expectation value of such operators with respect to a statevector $|\psi\rangle = \sum\limits_n^{2^N} \alpha_n |n\rangle$:

$$
\langle E \rangle = \langle\psi|\hat{H}|\psi\rangle = \sum\limits_i^T c_i \langle\psi| \hat{\sigma}_i |\psi\rangle 
\\,\\,\\,\\, \in \mathbb{R}.
$$

Evaluating each $\langle\psi| \hat{\sigma}_i |\psi\rangle$ is simple. Observe that:
- $\hat{\sigma}_i |\psi\rangle$ merely swaps pairs of (scaled) amplitudes (see page 24 of [here](https://arxiv.org/pdf/2311.01512)).
- precisely, the action on the $n$-th basis state is $` \hat{\sigma}_i |n\rangle = s_{i,n}|n'\rangle `$ where $` s_{i,n} \in \{\pm 1, \pm \mathrm{i}\} `$.
- the "swapped" index $n'$ is simply $n$ but with a bit flipped at the location of every $\hat{X}$ and $\hat{Y}$ operator in $\hat{\sigma}_i$.
- ergo $`\langle\psi| \hat{\sigma}_i |\psi\rangle =  \sum\limits_n^{2^N} \sum\limits_m^{2^N} \alpha_n {\alpha_m}^* \langle m | \hat{\sigma}_i |n\rangle = \sum\limits_n^{2^N} \sum\limits_m^{2^N} \alpha_n {\alpha_m}^* \langle m | s_{i,n} |n'\rangle = \sum\limits_n^{2^N} \alpha_n {\alpha_{n'}}^* s_{i,n}  `$

So each $\langle\psi| \hat{\sigma}_i |\psi\rangle$ is just an out-of-order (and strangely weighted) inner-product of $|\psi\rangle$ with itself! You can see its simple evaluation [here](https://github.com/QuEST-Kit/QuEST/blob/53f8f3ad60e5b0171646ee8250c0e6ea65878b44/quest/src/cpu/cpu_subroutines.cpp#L2059). In distributed settings, obtaining $` \alpha_{n'} `$ may require we first send all of a node's local amplitudes to a unique pair-node - we do this [here](https://github.com/QuEST-Kit/QuEST/blob/53f8f3ad60e5b0171646ee8250c0e6ea65878b44/quest/src/core/localiser.cpp#L1998). The code to evaluate $\langle\psi| \hat{\sigma}_i |\psi\rangle$ then boils down to:
```python
def calcExpecPauliStr(qureg, str, remote_amps):
    out = 0

    for (ind, amp) in enumerate(qureg):

        # index of other amp determined by locations of X or Y in str
        other_ind = getOtherAmpIndex(index, str)
        other_amp = remote_amps[other_ind]

        out += conj(amp) * other_amp * getStrPrefactor(ind,str)

    return out
```

How now should we evaluate the _sum_ $\langle\psi|\hat{H}|\psi\rangle$? 

### Unoptimised

To evaluate $\langle\psi|\hat{H}|\psi\rangle$, we _could_ naively evaluate each $\langle\psi| \hat{\sigma}_i |\psi\rangle$ in-turn and sum them, weighted by $c_i$.

The serial logic would resemble
```python
def calcExpecPauliStrSum(qureg, sum):
    out = 0

    for (coeff, str) within sum:
        out += coeff * calcExpecPauliStr(qureg, str)

    return out
```

while when `qureg` is distributed, we may need to pre-communicate amplitudes needed by `calcExpecPauliStr`.

```python
def calcExpecPauliStrSum(qureg, sum):
    out = 0

    for (coeff, str) within sum:

        remote_amps = communicateData(qureg, str.property)
        out += coeff * calcExpecPauliStr(qureg, str, remote_amps)

    return out
```

Notice that every separate `str` in `sum` may separately invoke inter-node communication (through `communicateData`). This turns out to be wasteful and _not_ what we currently do!

### Currently

The above in-turn processing is wasteful because many distinct $\hat{\sigma}_i \in \hat{H}$ require _identical_ communication patterns to simulate. This is ultimately due to multiple $\hat{\sigma}_i$ producing the same $n'$ from $n$. That is, $` \hat{\sigma}_i |n\rangle = s_{i,n}|n'\rangle `$ _and_ $` \hat{\sigma}_j |n\rangle = s_{j,n}|n'\rangle `$ for the same $|n'\rangle$. This is true whenever $\hat{\sigma}_i$ and $\hat{\sigma}_j$ differ only by the substitution of $\hat{X} \leftrightarrow \hat{Y}$ and $\hat{I} \leftrightarrow \hat{Z}$. For all amplitudes at indices $|n\rangle$ in a node (of rank $r$), all its corresponding amplitudes $|n'\rangle$ are contained in a single other node (of rank $r'$). This other rank $r'$ is determined (from $r$) entirely by the Pauli operators in $\hat{\sigma}$ which are in the "prefix partition" (see page 11 [here](https://arxiv.org/pdf/2311.01512)).

So, we instead group all $` \{\hat{\sigma}_i\} `$ together which differ only by $\hat{X} \leftrightarrow \hat{Y}$ and $\hat{I} \leftrightarrow \hat{Z}$ in the _prefix qubits_. All Pauli strings in this group have the same communication pattern; so we need only communicate _once_ per group!

This is performed within the [`localiser_statevec_calcExpecPauliStrSum()`](https://github.com/QuEST-Kit/QuEST/blob/53f8f3ad60e5b0171646ee8250c0e6ea65878b44/quest/src/core/localiser.cpp#L2050) function, and resembles logic:

```python
def calcExpecPauliStrSum(qureg, sum):
    out = 0

    groups = groupStringsByPrefixPaulis(sum)

    for group in groups:
        remote_amps = communicateData(qureg, group.property)
    
        for (coeff, str) in group:
            out += coeff * calcExpecPauliStr(qureg, str, remote_amps)

    return out
```

Grouping strings by their prefix Paulis in this way greatly reduces the amount of communication performed, significantly optimising distributed simulation. But with more thinking, we can accelerate even non-distributed simulation further!

### Further optimisation?

We _might_ further speed up calculating $\langle\psi|\hat{H}|\psi\rangle$ by further grouping strings. We saw $\hat{\sigma}_i$ and $\hat{\sigma}_j$, which differ only by $\hat{X} \leftrightarrow \hat{Y}$ and $\hat{I} \leftrightarrow \hat{Z}$, linearly combine the _same pair_ of amplitudes in the innermost sum. We can exploit this to avoid superfluous re-enumeration of all amplitudes, which contributes significantly to the total runtime due to statevector simulation typically being [memory bandwidth bound](https://en.wikipedia.org/wiki/Memory-bound_function).

We can do this by grouping same-communication strings into those with _suffix_ Paulis differing by $\hat{X} \leftrightarrow \hat{Y}, \hat{I} \leftrightarrow \hat{Z}$. Then, for each group, we enumerate all to-be-combined amplitude pairs, and for each pair, enumerate all Paulis in the group (informing the coefficient of the pair in the total sum). 

The logic may resemble:
```python
def calcExpecPauliStrSum(qureg, sum):
    out = 0

    groups = groupStringsByPrefixPaulis(sum)

    for group in groups:
        remote_amps = communicateData(qureg, group.property)

        subgroups = groupStringsBySuffixPaulis(group)

        for subgroup in subgroups:
            out += calcExpecPauliStrGroup(qureg, subgroup, remote_amps)

    return out
```
where
```python
def calcExpecPauliStrGroup(qureg, subgroup, remote_amps):
    out = 0

    for (ind, amp) in enumerate(qureg):

        # index of other amp determined by locations of X or Y
        # (which is the same for all strings within subgroup)
        other_ind = getOtherAmpIndex(index, subgroup[0])
        other_amp = qureg[other_ind]

        total = 0
        for str in subgroup:
            total += subgroup.coeffs[ind] * getStrPrefactor(ind,str)

        out += total * conj(amp) * other_amp 

    return out
```

This is _asymptotically_ equivalent to the previous semi-optimised method, but changes the "parallelisation granularity". Since the number of amplitudes greatly exceeds the number of Pauli strings (such that all Pauli strings and their coefficients fit in much fewer cache lines), this should reduce memory-fetch costs. This benefit is likely to outweigh the incurred slowdown from introducing branching in the hot loop (via the new inner for loop).

## Challenge details

The new routines should be divided between [`localiser.cpp`](https://github.com/QuEST-Kit/QuEST/blob/devel/quest/src/core/localiser.cpp) (to form the groups) and [`cpu_subroutines.cpp`](https://github.com/QuEST-Kit/QuEST/blob/devel/quest/src/cpu/cpu_subroutines.cpp) (to simulate a group).

> It is **not necessary** for _unitaryHACK_ to adapt the equivalent GPU functions in [`gpu_subroutines.cpp`](https://github.com/QuEST-Kit/QuEST/blob/devel/quest/src/gpu/gpu_subroutines.cpp)

In lieu of defining new functions therein however, you should adapt the existing routines, which are:
- [`localiser_statevec_calcExpecPauliStrSum()`](https://github.com/QuEST-Kit/QuEST/blob/53f8f3ad60e5b0171646ee8250c0e6ea65878b44/quest/src/core/localiser.cpp#L2050)
- [`cpu_statevector_anyCtrlPauliTensorOrGadget_subA()`](https://github.com/QuEST-Kit/QuEST/blob/53f8f3ad60e5b0171646ee8250c0e6ea65878b44/quest/src/cpu/cpu_subroutines.cpp#L2059)
- [`cpu_statevector_anyCtrlPauliTensorOrGadget_subB()`](https://github.com/QuEST-Kit/QuEST/blob/53f8f3ad60e5b0171646ee8250c0e6ea65878b44/quest/src/cpu/cpu_subroutines.cpp#L2090)

and the corresponding [`accelerator.cpp`](https://github.com/QuEST-Kit/QuEST/blob/53f8f3ad60e5b0171646ee8250c0e6ea65878b44/quest/src/core/accelerator.cpp#L990) functions (which will likely just need a trivial signature change). You can read about the software architecture [here](https://github.com/QuEST-Kit/QuEST/blob/devel/docs/architecture.md).

> It is **not necessary** for _unitaryHACK_ to derive a similar optimisation for the _density-matrix_ equivalent functions.

## Testing

For the solution to be _integrated_ into QuEST (but not necessarily for it to be accepted as UH solution), it will be necessary to verify both the correctness of the solution (in both local and distributed settings), _and_ its performance benefit. This will be guided be me in a followup.

luffy-orf:
This PR implements the batch-processing optimization for calcExpecPauliStrSum as described in issue #599.

- Groups Pauli strings by both prefix (communication pattern) and suffix (amplitude-mixing pattern).
- Processes all Pauli strings in a suffix group in a single pass over the statevector, reducing redundant memory access.
- Adds a static helper for suffix grouping.
- Refactors localiser_statevec_calcExpecPauliStrSum to use the new grouping and processing logic.
- Only applies this optimization to the statevector case, not density matrix.
- Adds comments to clarify the new logic.

Closes #599. 
Solved under unitaryhack

### DIFF
--- a/quest/src/core/localiser.cpp
+++ b/quest/src/core/localiser.cpp
@@ -23,6 +23,7 @@
 #include "quest/src/core/accelerator.hpp"
 #include "quest/src/comm/comm_config.hpp"
 #include "quest/src/comm/comm_routines.hpp"
+#include "quest/src/core/fastmath.hpp"
 #include "quest/src/cpu/cpu_config.hpp"
 #include "quest/src/gpu/gpu_config.hpp"
 
--- a/quest/src/core/localiser.cpp
+++ b/quest/src/core/localiser.cpp
@@ -32,6 +32,7 @@
 #include <complex>
 #include <algorithm>
 #include <unordered_map>
+#include <functional>
 
 using std::vector;
 using std::tuple;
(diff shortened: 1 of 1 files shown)


TysonRayJones:
Hi there,

This is a great start! Note that the multithreaded routine should be defined in [`cpu_subroutines.cpp`](https://github.com/QuEST-Kit/QuEST/blob/devel/quest/src/cpu/cpu_subroutines.cpp) with the other OpenMP functions (which is causing the current compilation issues). An equivalent GPU function is also needed in [`gpu_subroutines.cpp`](https://github.com/QuEST-Kit/QuEST/blob/devel/quest/src/gpu/gpu_subroutines.cpp), although it's not essential to prepare it for this challenge.

In lieu of defining new functions therein however, you should adapt the existing routines, which are:
- [`cpu_statevector_anyCtrlPauliTensorOrGadget_subA()`](https://github.com/QuEST-Kit/QuEST/blob/53f8f3ad60e5b0171646ee8250c0e6ea65878b44/quest/src/cpu/cpu_subroutines.cpp#L2059)
- [`cpu_statevector_anyCtrlPauliTensorOrGadget_subB()`](https://github.com/QuEST-Kit/QuEST/blob/53f8f3ad60e5b0171646ee8250c0e6ea65878b44/quest/src/cpu/cpu_subroutines.cpp#L2090)

and the corresponding [`accelerator.cpp`](https://github.com/QuEST-Kit/QuEST/blob/53f8f3ad60e5b0171646ee8250c0e6ea65878b44/quest/src/core/accelerator.cpp#L990) functions (which will likely just need a trivial signature change). You can read about the software architecture [here](https://github.com/QuEST-Kit/QuEST/blob/devel/docs/architecture.md).

You can ignore all density-matrix variants. I'll copy this info into the challenge issue when I get a moment!

EDIT: please PR also to _devel_ rather than main (I tried to re-base it by it duplicated some commits).

Issue #594, opened 2025-04-27
Extend Pauli rotations to accept complex angles


TysonRayJones:
## Summary

Implement `applyNonUnitaryPauliGadget()` which extends  [`applyPauliGadget()`](https://quest-kit.github.io/QuEST/group__op__pauligadget.html#gaf9c0090d2d98918743a6d713dabbd16d) to accept a _complex_ angle parameter. This requires merely updating several signatures in [`operations.cpp`](https://github.com/QuEST-Kit/QuEST/blob/devel/quest/src/api/operations.cpp) and [`localiser.cpp`](https://github.com/QuEST-Kit/QuEST/blob/devel/quest/src/core/localiser.cpp).

## Context

The existing function [`applyPauliGadget()`](https://quest-kit.github.io/QuEST/group__op__pauligadget.html#gaf9c0090d2d98918743a6d713dabbd16d) accepts a real scalar $\theta=$ `angle` and a `PauliStr` $\hat{\sigma}=$ `str` (a tensor product of Pauli operators), and simulates the unitary operation

$$
\hat{U}(\theta) = \exp\left( -\mathrm{i} \frac{ \theta}{2} \hat{\sigma} \right),  \\,\\,\\,\\,\\,\\,\\,\\,\\, \theta \in \mathbb{R}
$$

This is a multi-qubit generalisation of a [rotation operator](https://www.quantum-inspire.com/kbase/rotation-operators/) and appears in _many_ quantum circuits, like Trotter simulations.

Consider the _non_-unitary produced by substituting the real $\theta$ with an imaginary scalar, which removes the imaginary unit from the exponent:

$$
\hat{V}(\theta) = \hat{U}(- \theta \mathrm{i}) =  \exp\left( - \frac{ \theta}{2} \hat{\sigma} \right),  \\,\\,\\,\\,\\,\\,\\,\\,\\, \theta \in \mathbb{R}.
$$

Despite being non-physical, such an operation turns out to be very useful in a classical simulator. It appears in Trotter circuits for "[imaginary time evolution](https://physics.stackexchange.com/questions/557225/why-do-we-use-the-imaginary-time-evolution-in-simulations-of-some-quantum-system)" which when effected upon a state, drives the system toward the ground-state. 

A slightly more general variant of the operation would permit the input parameter to be _any_ complex number.

$$
[...]

PR #632, opened 2025-05-29
Extend Pauli rotations to accept complex angles #594 (unitaryhack2025)

diogomaia00:
In order to resolve the issue of the unitary hackathon:
- Changes were made in the files (localiser.cpp, utilities.cpp e operations.cpp) 
- Test case in the examples/automated/apply_no_unitary_pauli_gadget.c

PR #633, opened 2025-05-29
v2 Extend Pauli rotations to accept complex angles #594 (unitaryhack2025)

diogomaia00:
Corrections made, sorry for the problems with the whitespaces


TysonRayJones:
Thanks for following up from #632! Alas you can see some compilation is failing - you've forgotten to expose the new operation in the `operations.h` header. You can double-check everything that's needed by (in your editor of choice) searching for all matches for `applyPhaseGadget()`, which remains a corresponding function.

Note to merge your PR, it would be necessary to delete the previous, redundant code (rather than leave it merely commented out), and avoid duplicating the previous commits by _forking_ from `devel` (rather than from `main`)

diogomaia00:
I will just do another PR with all the correct steps you’ve mentioned

PR #645, opened 2025-06-11
Fix for quest.h generation

otbrown:
Hi Tyson,

Hope the unitary hack is going well!

I've finally updated to 4.1 locally and spotted a subtle bug with Luc's custom `compile_option` function in CMake -- the variables set within it were dutifully thrown away at return. The fix is trivial, just needed to added `PARENT_SCOPE` to the set command.

As that function also independently set `target_compile_definitions` the library itself is still compiled correctly, this bug would only impact downstream projects (like my CQ!).

Cheers,
Oliver

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,7 +79,7 @@ else()
   set(MULTI_LIB_HEADERS 0)
   function(compile_option VAR VALUE)
     target_compile_definitions(QuEST PRIVATE ${VAR}=${VALUE})
-    set(${VAR} ${VALUE})
+    set(${VAR} ${VALUE} PARENT_SCOPE)
   endfunction()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -576,13 +576,18 @@ install(FILES
         DESTINATION "${QuEST_INSTALL_CONFIGDIR}"
 )
 
-install(FILES "${CMAKE_CURRENT_BINARY_DIR}/include/quest.h"
+install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/quest/include/quest.h"
         DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
+)
+
+install(FILES "${CMAKE_CURRENT_BINARY_DIR}/include/quest/include/config.h"
+        DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/quest/include"
 )
 
 install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/quest/include"
         DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/quest"
         FILES_MATCHING PATTERN "*.h"
+        PATTERN "quest.h" EXCLUDE
 )
 
 install(EXPORT QuESTTargets
--- a/quest/include/CMakeLists.txt
+++ b/quest/include/CMakeLists.txt
@@ -2,4 +2,4 @@
 # @author Erich Essmann
 # @author Luc Jaulmes (using config file)
 
-configure_file(quest.h.in "${CMAKE_BINARY_DIR}/include/quest.h" @ONLY)
+configure_file(config.h.in "${CMAKE_BINARY_DIR}/include/quest/include/config.h" @ONLY)
--- a/quest/include/config.h.in
+++ b/quest/include/config.h.in
@@ -1,6 +1,12 @@
 #ifndef CONFIG_H
 #define CONFIG_H
 
+// be warned, the below is sensitive to whitespace after the slash
+#if !defined(FLOAT_PRECISION)\
+    || !defined(COMPILE_MPI)\
+    || !defined(COMPILE_OPENMP)\
+    || !defined(COMPILE_CUDA)\
+    || !defined(COMPILE_CUQUANTUM)
 
 // bind compile settings to installed exec
 #if !@MULTI_LIB_HEADERS@
@@ -11,5 +17,6 @@
 #cmakedefine01 COMPILE_CUQUANTUM
 #endif
 
+#endif
 
 #endif
--- a/quest/include/config.h.in
+++ b/quest/include/config.h.in
@@ -1,0 +1,15 @@
+#ifndef CONFIG_H
+#define CONFIG_H
+
+
+// bind compile settings to installed exec
+#if !@MULTI_LIB_HEADERS@
+#cmakedefine FLOAT_PRECISION @FLOAT_PRECISION@
+#cmakedefine01 COMPILE_MPI
+#cmakedefine01 COMPILE_OPENMP
+#cmakedefine01 COMPILE_CUDA
+#cmakedefine01 COMPILE_CUQUANTUM
+#endif
+
+
+#endif
--- a/quest/include/modes.h
+++ b/quest/include/modes.h
@@ -15,27 +15,8 @@
 
 
 
-// ensure all mode flags are defined
-
-#ifndef COMPILE_MPI
-    #error "Compiler must define COMPILE_MPI"
-#endif
-
-#ifndef COMPILE_OPENMP
-    #error "Compiler must define COMPILE_OPENMP"
-#endif
-
-#ifndef COMPILE_CUDA
-    #error "Compiler must define COMPILE_CUDA"
-#endif
-
-#ifndef COMPILE_CUQUANTUM
-    #error "Compiler must define COMPILE_CUQUANTUM"
-#endif
-
-
-
 // ensure all mode flags are valid values
+// undefined allowed as undefined == 0 in C/C++ standards
 
 #if ! (COMPILE_MPI == 0 || COMPILE_MPI == 1)
     #error "Macro COMPILE_MPI must have value 0 or 1"
--- a/quest/include/quest.h
+++ b/quest/include/quest.h
@@ -31,19 +31,11 @@
 #define QUEST_H
 
 
-// bind compile settings to installed exec
-#if !@MULTI_LIB_HEADERS@
-#cmakedefine FLOAT_PRECISION @FLOAT_PRECISION@
-#cmakedefine01 COMPILE_MPI
-#cmakedefine01 COMPILE_OPENMP
-#cmakedefine01 COMPILE_CUDA
-#cmakedefine01 COMPILE_CUQUANTUM
-#endif
-
-
 // include version first so it is accessible to 
 // debuggers in case a subsequent include fails
 #include "quest/include/version.h"
+
+#include "quest/include/config.h"
 
 // include before API headers since it validates
 // preprocessor configuration, and affirms macro


otbrown:
Hmm I'm now getting hit by a bunch of multiple definition errors when attempting to compile CQ-SimBE as a consequence of trying to include `quest.h` in a header file, fair enough. If I _don't_ include `quest.h` though I fail a `modes.h` check. 

All of which is to say I think I'm now an advocate of separatiing the `COMPILE_XXX` definitions into a `quest_config.h`!

TysonRayJones:
Drat! It's true that those preprocessors aren't _needed_ by the user source; they only affect backend compilation. They need only "survive" past compilation for users who wish to know the compiled backends during compilation of their own source code. So I'm with you about moving them to a bespoke, optional header.

Then `quest.h.in` would return to `quest.h`, and we could adapt `modes.h` from
```C
// ensure defined
#ifndef COMPILE_MPI
    #error "Compiler must define COMPILE_MPI"
#endif

// ensure valid
#if ! (COMPILE_MPI == 0 || COMPILE_MPI == 1)
    #error "Macro COMPILE_MPI must have value 0 or 1"
#endif
```
to
```C
// permit undefined else ensure valid
#if ! (COMPILE_MPI == 0 || COMPILE_MPI == 1)
    #error "Macro COMPILE_MPI must have value 0 or 1, or else be undefined"
#endif

// (above exploits undefined preprocessor == 0 in C/C++ standards)
```
and rename it to `preprocessors.h` or `integrity.h` - or something to indicate it merely validates optionally-defined preprocessors.

The new `quest_config.h` you propose would then contain Luc's cmake-fu, and need only ever be included by "superusers" who demand compile-time knowledge of the compiled backends. It can itself be defensive and include guards to ensure `COMPILE_XXX` are _not_ defined before its inclusion.

So all agreed! Alas I'm insufficiently knowledgeable in CMake configs to make `quest_config.h.in` properly, but happy to help out alongside e.g. yourself or @lucjaulmes 🙏

otbrown:
I think the changes needed are pretty minor, so will try them out over the next few days and hit you with the PR! If I get stuck I'll definitely be asking Luc ;)

otbrown:
Okay! We no longer generate `quest.h`, but instead generate `quest/include/config.h`, for inclusion if needed by ~~problem~~ advanced users like me ;)

TysonRayJones:
Wew brilliant! :tada:

## Enabling non-cmake compilation

Can I throw a minor spanner in the works? 😅 It'd be neat to make extricable the build from cmake so that manual compilation (like through [this bash script](https://github.com/QuEST-Kit/QuEST/blob/main/utils/scripts/compile.sh) - though I must update it) remains possible. It's inessential but super useful for debugging compilation.

This is currently only prevented by `quest.h` containing:
```C++
#include "quest/include/config.h"
``` 
since `config.h` does not exist until CMake processes `config.h.in`. Can we adapt it to avoid seeking `config.h` when CMake isn't used?

Some quick ideas:
  1. guard inclusion of `config.h` from within `quest.h` by some preprocessor being _defined_.
     Then users of `quest.h` never see `COMPILE_XXX` _unless_ they define e.g. `DEFINE_CONFIG` (and they compiled or installed with cmake, of course) _before inclusion_. Would not affect cmake config generation at all.
     > cons: some of the API (arguably unimportant compile-time stuff) is then "hidden by default"
  2. guard inclusion of `config.h` from within `quest.h` by some preprocessor being _**not** defined_. Then "manual compilers" (pejorative) must define e.g. `DO_NOT_DEFINE_CONFIG` to successfully compile.
     > cons: manual compilers _must_ define that macro to compile. This is a bit of a tedium in the intended debugging settings, complicating the compiler arguments which are the very things being scrutinised.
  3. remove `config.h` from `quest.h` and require users directly include it (renamed to e.g. `quest_config.h`).
     > cons: it's a shame to need multiple includes - and does having such muddy the installation process?
  4.  rename `config.h.in` to `config.h`, guard against the `#cmakedefine` macros using a cmake-defined preprocessor, configure cmake to process `config.h` into `config_out.h` which is included at the bottom of `config.h`. So `config.h` prior exists and is always included, and will itself include a secondary cmake-generated file with the permanent macros.
      > cons: seems hacky and nonstandard
   
Do you have any thoughts on these? I lean personally toward **1.** with the justification that `COMPILE_XXX` are outside the main API; they are compile-time and ergo "special", so warrant needing to use a link-time macro to observe.


## Other preprocessors

The `DEFAULT_VALIDATION_EPSILON` macro in `precision.h` is also advertised as re-configurable at compile-time. Since it is not included in the cmake config, it is possible for users to override it to different values when installing then when including QuEST. Alas changing `DEFAULT_VALIDATION_EPSILON` from the value used during installation does _nothing_. Worse, users might override it at installation but then _not_ override it during inclusion (which is natural) but then consult its value which will be wrong.

This does not seem a _huge_ problem: it's "too late" for users to adjust the default epsilon, and they'll instead have to use [runtime control](https://quest-kit.github.io/QuEST/group__debug__validation.html#gae395568df6def76045ec1881fcb4e6d1). This might be better than caching `DEFAULT_VALIDATION_EPSILON` by moving it out of `precision.h` (where it needs to consult `FLOAT_PRECISION`) into `config.h.in`. It does however necessitate updating the doc...

https://github.com/QuEST-Kit/QuEST/blob/f3baf3455956c03f98f351e69bcdb5d47b45562b/quest/include/precision.h#L123-L131

to e.g.
```C++
/*
 * RE-CONFIGURABLE DEFAULT VALIDATION PRECISION
 *
 * which is overridable when compiling the QuEST source by defining 
 * DEFAULT_VALIDATION_EPSILON. Beware that updating this macro
 * after compilation (e.g. after installing QuEST) has no effect, though
 * epsilon can be overridden at runtime using setValidationEpsilon()
 */
```
 
What do you think about this? Is it a pitfall? Will users deliberately exposing the config macros (as per above) be tricked by `DEFAULT_VALIDATION_EPSILON` (provided by a different header) being distinct from its install-time value which is actually being used by the source?

otbrown:
RE non-cmake compilation:

So actually, with the removal of the modes header `#ifndef COMPILE_XXXX` guards, the _only_ thing that doesn't compile if you just remove [`#include "quest/include/config.h"`](https://github.com/otbrown/QuEST/blob/1f1e70288a57beca091f8f84bc55eefdad037ac4/quest/include/quest.h#L38) is [one line in the tests main](https://github.com/otbrown/QuEST/blob/1f1e70288a57beca091f8f84bc55eefdad037ac4/tests/main.cpp#L95) where it checks if cuQuantum is compiled. In fact, even my downstream project doesn't need those compiler definitions now that guarding is removed, so I vote for option 3 -- remove it from quest.h. We should still generate it, as it's nice to have a record of what was compiled, but that's now its main purpose! Users who are treating QuEST strictly as a library no longer need to know or care about these defines, which feels like a win all round. 

I would of course also update `QuESTEnv` to include `int isCuQuantum` or equivalent to remove the problem line in tests/main.cpp too.

RE `DEFAULT_VALIDATION_EPSILON`: 

To be honest I suspect a user messing with this _really_ needs to know what they're doing, so I'm not too upset about it being fiddly to do :) Here I vote for just updating the doc as you suggest to make it clear it's a property of the QuEST library, not of the application code they build against that library.

otbrown:
I can see an argument here for expanding config.h to include the default validation epsilon set at compile time, for similar record-keeping/occasional-onward-propagation purposes!

TysonRayJones:
### `COMPILE_CUQUANTUM`

Aha I'd forgotten about that reference in the tests - I'll bind `isCuQuantumEnabled` to `QuESTEnv` as suggested :pray: Then agreed `config.h` (or renamed to `quest_config.h` for standalone inclusion) can be removed from `quest.h`.

### `DEFAULT_VALIDATION_EPSILON`

You've convinced me that `quest_config.h` _should_ include `DEFAULT_VALIDATION_EPSILON`. Will the below "default logic" have to be moved out of the headers and into the CMake build? 

https://github.com/QuEST-Kit/QuEST/blob/4d44ec81debdae20cb6ae988e400bf606180a65e/quest/include/precision.h#L133-L140

### `PERMIT_NODES_TO_SHARE_GPU`

> TLDR I will change `PERMIT_NODES_TO_SHARE_GPU` to an environment variable unless objected to! Reasoning below (in context) for posterity.

The remaining preprocessors which can differ between compilation and include are:

https://github.com/QuEST-Kit/QuEST/blob/4d44ec81debdae20cb6ae988e400bf606180a65e/quest/include/modes.h#L78-L88

The latter two are unimportant and _should_ be link-time overridable since they merely control user-source-compile-time warnings. But `PERMIT_NODES_TO_SHARE_GPU` is stickier - it appears it two sources:
https://github.com/QuEST-Kit/QuEST/blob/4d44ec81debdae20cb6ae988e400bf606180a65e/quest/src/api/environment.cpp#L105-L111

https://github.com/QuEST-Kit/QuEST/blob/4d44ec81debdae20cb6ae988e400bf606180a65e/quest/src/core/validation.cpp#L1361-L1370

> (Its second appearance is totally redundant; that must be that 3am flavour of "defensive design" woops 😅 )

The purpose of `PERMIT_NODES_TO_SHARE_GPU` is just to disable validation which prevent users from mistakingly over-subscribing GPUs and experiencing unexpected performance degradation. It's useful to disable this validation when unit-testing MPI+GPU hybridisation on a single GPU.

If it were to remain a compile-time flag, I suppose it should also be recorded by `quest_config.h`. _BUT_ there's really no need for this to be fixed at compile-time - it seems a gratuitous tedium to have to recompile/re-install QuEST just to change it! Ideally one should wish to easily change it between launches of a QuEST executable.

It's a little clumsy to be specified at runtime however, since that validation occurs during `initQuESTEnv()` before which _no_ user-configuration can be done. It also seems gross to add it as a necessary parameter to [`initCustomQuESTEnv()`](https://quest-kit.github.io/QuEST/group__environment.html#ga485268e52f838743357e7a4c8c241e57) since it "feels" like a meta setting that should never need to be referenced except when being disabled.

So, it seems prudent to change it to an environment-variable. Would this cause any issues? Would that jeopardise QuEST's ability to run on extremely stripped-down systems like microcontrollers?? 😅 

### Test macros

Is it okay to _neglect_ the test-specific macros? Are the tests even included in the installation?

https://github.com/QuEST-Kit/QuEST/blob/4d44ec81debdae20cb6ae988e400bf606180a65e/tests/utils/macros.hpp#L30-L48

TysonRayJones:
I've had the :shower: epiphany :shower: that the discussed overridable preprocessors fall into three categories: those which must be known at...
  1. source-compile-time (e.g. `COMPILE_XXX`)
  2. user-source-link-time (e.g. `INCLUDE_DEPRECATED_FUNCTIONS`)
  3. executable launch, before `initQuESTEnv()` (e.g. `PERMIT_NODES_TO_SHARE_GPU`)

All 1. must be saved by cmake, all 2. must never be saved, and 3. should never have been preprocessors! Instead, all 3. should be _environment variables_; they offer no benefit to being decided at source-compile-time which instead reduces flexibility. I propose changing all below macros to environment variables:
_(source)_
  - `PERMIT_NODES_TO_SHARE_GPU`
  - `DEFAULT_VALIDATION_EPSILON`

_(tests)_
  - `TEST_MAX_NUM_QUBIT_PERMUTATIONS`
  - `TEST_MAX_NUM_SUPEROP_TARGETS`
  - `TEST_ALL_DEPLOYMENTS`
  - `TEST_NUM_MIXED_DEPLOYMENT_REPETITIONS`

The latter four are a little irritating to refactor since they are accessed outside the source so shouldn't use the internal convenience functions for obtaining env-vars. This puts 'don't repeat yourself' in tension with 'separation of concerns', but I'll find some compromising design.

I'll ergo not merge #650 and instead extend it in a new PR to properly handle environment variables. That should resolve _everything_; there will be no remaining preprocessors needing saving by cmake, no ineffectual-overridable macros, `quest_config.h` can be strictly "include as needed" (necessitating cmake compilation), yet the source can remain manually compilable. 👨‍🍳 :kiss:

otbrown:
Sounds great to me!

(Apologies for the delay, was travelling now attending a conference!)

TysonRayJones:
Wew brilliant! No problem about the micro-delay - I'm supposed to be on holiday but I have norovirus 😭 

I'll merge this PR, commit the env-vars, then will commit renaming of `quest_config.h` to `quest_config.h.in` and add some file doc. 

One final thing regarding these lines:

https://github.com/QuEST-Kit/QuEST/blob/1f1e70288a57beca091f8f84bc55eefdad037ac4/quest/include/config.h.in#L5-L9

~~Is it correct that this permits gratuitous specification of `COMPILE_XXX` so that e.g. compiling _without installing_ will work? Is it plausible a user could install QuEST but later erroneously believe they can change the available hardware backends using those preprocessors (and also be including `quest_config.h` for some reason)? That wouldn't cause any errors, but may cause [user astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) - though they can runtime confirm which deployments are actually compiled (using `reportQuESTEnv()`).~~

A user optionally including `quest_config.h` is seeking to _obtain_ `COMPILE_XXX` rather than override, so it seems sensible to actually throw a compile-time error if they are gratuitously specifying those preprocessors. I can chuck in the error checks in the aforementioned commit 🙏

otbrown:
Ooft, really sorry to hear about the norovirus 😬 that's a bad time even if you weren't meant to be on holiday...

Exactly so, this tells the user what was compiled into the QuEST library -- redefining elsewhere will, at best, do nothing. So not a bad idea to warn the user they're being silly. My original logic there was just that they all needed to be defined in the QuEST build, so if any was undefined they should all be defined.

Final question, should we rename them to `QUEST_COMPILE_XXXX` in case downstream codes have similar ideas around compile time configuration? It seems _fairly_ unlikely to clash, especially now that it's not leaking onward at all, but still...

TysonRayJones:
Thanks very much! It's pretty grim, though it's making me appreciate all the good work my bowels ordinarily perform.

Gotcha! A `QUEST_` prefix seems sensible and standard - I suppose _all_ preprocessors (and maybe even environment variables?!) should be prefixed. Does changing preprocessors constitute an API break needing a major version update? 😓

otbrown:
I would say no, as user code stays the same! (Unless you were doing something _really_ weird). It only maybe matters to whoever builds the library.

Plus most people will set these preprocessor directives through the magic of CMake, and that interface remains unchanged 😁

lucjaulmes:
Sorry this is a bit verbose and I’m joining a bit late. But everything seems sorted now, no? :sweat_smile:

TysonRayJones:
Ok perf! Let's rename the vars in another PR (after I merge the env-var one) to make that change more explicit. Sorry for the delay in merging this!

PR #649, opened 2025-06-13
bound isCuQuantumEnabled to QuESTEnv

TysonRayJones:
so that the COMPILE_CUQUANTUM preprocessor need only ever be consulted by the source during compilation, as per Oliver's suggestion in #645

### DIFF
--- a/quest/include/environment.h
+++ b/quest/include/environment.h
@@ -32,10 +32,13 @@ extern "C" {
 /// @notyetdoced
 typedef struct {
 
-    // deployment mode
+    // deployment modes which can be runtime disabled
     int isMultithreaded;
     int isGpuAccelerated;
     int isDistributed;
+
+    // deployment modes which cannot be directly changed after compilation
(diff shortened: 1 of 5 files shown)


TysonRayJones:
For posterity: one might wonder whether it's natural or worthwhile to support _runtime_ or _post-compilation_ toggling of cuQuantum. That is:
- cuQuantum remains compile-time optional so that use of QuEST does not require its installation
- if cuQuantum was compiled, it can be runtime disabled using e.g. `initCustomQuESTEnv()`, or an environment variable.
- if it was not compiled, attempting to runtime force-enable it would trigger an error message

The benefits of this are not very compelling. It would enable performance comparisons of cuQuantum vs QuEST kernels without recompiling the QuEST source. That's a niche usecase which should be made redundant by _us_ only using cuQuantum where performant. This benefit does not outweigh the necessary increase of the code complexity in `gpu_subroutines.cpp` to perform both compile-time _and_ runtime branching.

PR #650, opened 2025-06-14
changed PERMIT_NODES_TO_SHARE_GPU from macro to env-var

TysonRayJones:
This allows it to be changed post-compilation/installation, pre-execution, as per the machinations in #645

Additionally inserted whitespaces into cmake error message about MacOS multithreading to make the advised commands clearer


TysonRayJones:
superseded by #653

PR #659, opened 2025-07-10
patched broken CUDA state after malloc failure

TysonRayJones:
Danny Hindson may have discovered a bug wherein the failing `cudaMalloc()` call deliberately induced by the 'out-of-memory' unit test of `createFullStateDiagMatr()` breaks subsequent GPU simulation. This could be explained by a failing `cudaMalloc` corrupting the CUDA API state until being explicitly cleared using the undocumented facility of `cudaGetLastError` (which clears "non-sticky' errors, see here). We correct this, and defensively check for irrecoverable sticky errors.

### DIFF
--- a/quest/src/core/errors.cpp
+++ b/quest/src/core/errors.cpp
@@ -698,6 +698,11 @@ void error_cudaCallFailed(const char* msg, const char* func, const char* caller,
     raiseInternalError(err);
 }
 
+void error_cudaEncounteredIrrecoverableError() {
+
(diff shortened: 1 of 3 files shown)


DannyHindson:
I've tested this on our HPC machines and it has fixed the problem. As discussed it was possible to see the problem by running the unit test suite with this filter applied:

```
build/tests/tests "applyPauliStr,createFullStateDiagMatr"
```

Thanks for looking into this quickly!

PR #668, opened 2025-07-22
renaming multiply functions

TysonRayJones:
This is a premature PR just to run unit-tests while actual renames are deliberated. It is now easier to all-replace `preapply` and `postapply` than it was to replace `multiply` and `postMultiply` (since the former collided with generic usage in comments)


TysonRayJones:
Now considering:
```C++
applyX();
leftapplyX();
rightapplyX();
```

Issue #638, opened 2025-06-03
Investigate potential single-CPU performance regression in v4


TysonRayJones:
In [v4](https://github.com/QuEST-Kit/QuEST/blob/main/docs/v4.md), the core algorithm of practically all QuEST routines fundamentally changed as per [this manuscript](https://arxiv.org/abs/2311.01512). As a comparison, compare this v3.6 routine:

https://github.com/QuEST-Kit/QuEST/blob/cc192644121cc7ac82e41630640bb69ea799627b/QuEST/src/CPU/QuEST_cpu.c#L2196-L2262

to its v4 equivalent:

https://github.com/QuEST-Kit/QuEST/blob/4d44ec81debdae20cb6ae988e400bf606180a65e/quest/src/cpu/cpu_subroutines.cpp#L361-L390

The v4 optimisations are focused upon avoiding redundant enumeration, avoiding branching within hot loops, and replacing integer arithmetic (which informs updated amplitude indices) with bitwise operations. The code was also cleaned to remove default and superfluous OpenMP boilerplate. As a consequence, the _order of accessed amplitudes_ in the state has substantially changed; from conditional albeit contiguous to unconditional but potentially unpredictable (by a compiler or branch-predictor). 

These changes are a benefit to GPU simulation which is detrimentally affected by warp-level looped branching. Alas, despite isolated testing during development, some changed functions may have regressed CPU performance as per #631 (and potentially #616). This _might_ be remedied in multithreaded multi-NUMA-node settings by distributing over nodes, or impending memory-pinning. But in any case, the potential damage to single-CPU performance remains.

**TODO**: get to the bottom of this! A quick isolated test shows v4 bitwise non-branched enumeration still beats v3's logic (at least on my Apple M1 Max). Figure out which functions have regressed and _why_!

TysonRayJones:
# Update 

A ~5x slowdown seen in serial functions (on my 2017 Macbook) like `applyRotateX` and `applyRotateY` - and here's why!

In v3, functions `rotateX`/`Y` wrapped the `compactUnitary` function, passing coefficients $\alpha$, $\beta$, which saw the simulation of matrix:

$$
e^{- \theta/2 i \sigma} \equiv
\begin{pmatrix}
\alpha & -\beta^* \\
\beta & \alpha^*
\end{pmatrix}.
$$

The logic resembles 

```cpp
// params of index algebra
long long int numTasks= qureg.numAmpsPerChunk / 2;
[...]

PR #684, opened 2025-09-08
patching MacOS performance

TysonRayJones:
as per #638

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -382,33 +382,65 @@ else()
   target_compile_definitions(QuEST PRIVATE INCLUDE_DEPRECATED_FUNCTIONS=0)
 endif()
 
-# Patch Clang/LLVM performance.
-# std::complex arithmetic overloads are painfully slow in Clang/LLVM, and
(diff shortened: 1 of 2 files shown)


TysonRayJones:
The below quick-and-dirty script did not reveal an accuracy difference when using `-Ofast` in Clang:

```cpp
int main() {
    initQuESTEnv();
    setMaxNumReportedSigFigs(20);

    // start in |+>
    Qureg qureg = createQureg(20);
    initPlusState(qureg);

    qreal angle = 0.3;

    // repeatedly effect R(a)R(-a) = I
    for (int q=0; q<qureg.numQubits; q++) {
        applyRotateX(qureg, q,   angle);
        applyRotateX(qureg, q, - angle);
    }

    // we should return to |+> with prob=0.5
    qreal prob = calcProbOfQubitOutcome(qureg, 0, 0);
    reportScalar("prob", prob);

    destroyQureg(qureg);
    finalizeQuESTEnv();
    return 0;
}


// -Ofast: prob = 0.49999999999999994449
// else:   prob = 0.49999999999999994449 (unchanged!)
```

PR #697, opened 2025-10-24
Simplify installation path configuration

### LINKED ISSUE #696: Stop prepending quest to the cmake install path
We should stop prepending the quest to the CMAKE_INSTALL_PREFIX, as altering any CMAKE variables is considered bad practice.

eessmann:
Removed unnecessary path normalization and appending for installation.
closes #696

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,8 +40,6 @@ set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake" ${CMAKE_MODULE_PATH})
 
 
 # GNUInstallDirs to provide sensible default install directory names
-cmake_path(SET ORG_INSTALL_PATH NORMALIZE "${CMAKE_INSTALL_PREFIX}")
-cmake_path(APPEND CMAKE_INSTALL_PREFIX "quest")
 include(GNUInstallDirs)
 include(CMakePackageConfigHelpers)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,10 +72,11 @@ endif()
 
 
 # Library type
-# Shared library by default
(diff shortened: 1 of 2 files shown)


otbrown:
Thanks for this Erich, I have two minor requests before merge:
- Not related to your changes but I see line 578 of CMakeLists.txt says `## RATH` instead of `## RPATH`!
- Could you change the executable we always build to be `quest-min-example` rather than `min_example`. If no CMAKE_INSTALL_PREFIX is set this will now end up in `/usr/local/bin` so we should namespace it...

otbrown:
Making the binaries specifically still be prepended by `quest/` would also be fine I think, although is possibly more faff! Though actually I'm now realising that all the examples may end up dumped in `/usr/local/bin` so we probably do want to make sure a `quest` gets inserted in there 🤔

eessmann:
I don't think we should install the executables, they are all just examples.

otbrown:
Good point. I was pondering over the weekend, and I can think of one or two awkward workflows where the user might want binaries installed too (i.e. when the user sets an install prefix and is using the USER_SOURCE option), so rather than remove the binary install entirely I was thinking ye olde create a CMake option. 

`option(INSTALL_BINARIES "Whether to include example and user binaries in the install." OFF)`  means the default behaviour is unsurprising, but people are supported in being weird without having to modify the CMake themselves.

eessmann:
Oliver, I’ve added your suggested option so we’re now conditionally installing the examples. I’ve also made it so that if we’re the top-level CMake project, we set the type of library we build to shared.

I encountered a bug when using Quest with the CMake FetchContent feature; Quest’s BUILD_SHARED_LIBS setting leaked into my build.

otbrown:
Excellent, thanks Erich!

PR #699, opened 2025-10-24
Thrust Overflow Fix

### LINKED ISSUE #698: thrust::make_counting_iterator(0) overflow
When targeting grace-hopper GPUs saw errors when trying to maximise size of problem that would fit on a Hopper GPU:

```
./qft 31
...
Total number of gates: 496
Measured probability amplitude of |0..0> state: 4.65661e-10
Calculated probability amplitude of |0..0>, C0 = 1 / 2^31: 4.65661e-10
terminate called after throwing an instance of 'std::runtime_error'
  what():  Invalid input range, passed negative size
Aborted (core dumped)
```
GPU should be able to fit 32 qubit calculations but limited to 30 in our initial tests.

Error related to overflows due to how `thrust::make_counting_iterators` are being used within `quest/src/gpu/gpu_thrust.cuh`.

Specifically functions:

`thrust_statevec_multiQubitProjector_sub`

`thrust_statevec_calcProbOfMultiQubitOutcome_sub`

However other instances of this clearly visible in file.

Initially fixed by updating the argument of `thrust::make_counting_iterator(0)` to `thrust::make_counting_iterator(0LL)`.

Will look to create pull request with a more comprehensive fix shortly.

otbrown:
gpu_thrust.cuh: modified initial thrust counting iterator declarations to use long long to avoid overflow at >30 qubits. Fixes #698.

### DIFF
--- a/quest/src/core/bitwise.hpp
+++ b/quest/src/core/bitwise.hpp
@@ -28,7 +28,9 @@
  */
 
 
-#define QINDEX_ONE 1ULL
+// alternatives to type-unsafe literals 
+constexpr qindex QINDEX_ZERO = 0; // used by gpu_thrust.cuh
+constexpr qindex QINDEX_ONE  = 1; // used only here
 
 
 INLINE qindex powerOf2(int exponent) {
--- a/quest/src/gpu/gpu_thrust.cuh
+++ b/quest/src/gpu/gpu_thrust.cuh
@@ -665,7 +665,7 @@ void thrust_fullstatediagmatr_setElemsToPauliStrSum(FullStateDiagMatr out, Pauli
         rank, out.numElems, logNumElemsPerNode,
         in.numTerms, toCuQcomps(out.gpuElems), devCoeffsPtr, devStringsPtr);
 
-    auto indIter = thrust::make_counting_iterator(0LL);
+    auto indIter = thrust::make_counting_iterator(QINDEX_ZERO);
     auto endIter = indIter + out.numElemsPerNode;
     thrust::for_each(indIter, endIter, functor);
 }
@@ -708,7 +708,7 @@ void thrust_densmatr_setAmpsToPauliStrSum_sub(Qureg qureg, PauliStrSum sum) {
         qureg.rank, powerOf2(qureg.numQubits), qureg.logNumAmpsPerNode,
         sum.numTerms, toCuQcomps(qureg.gpuAmps), devCoeffsPtr, devStringsPtr);
 
-    auto indIter = thrust::make_counting_iterator(0LL);
+    auto indIter = thrust::make_counting_iterator(QINDEX_ZERO);
     auto endIter = indIter + qureg.numAmpsPerNode;
     thrust::for_each(indIter, endIter, functor);
 }
@@ -771,7 +771,7 @@ qreal thrust_densmatr_calcTotalProb_sub(Qureg qureg) {
     /// reduction may be suboptimal; it may be necessary to
     /// implement a custom numerically-stable CUDA reduction.
 
-    auto rawIter = thrust::make_counting_iterator(0LL);
+    auto rawIter = thrust::make_counting_iterator(QINDEX_ZERO);
     auto indIter = thrust::make_transform_iterator(rawIter, functor_getDiagInd(qureg));
     auto ampIter = thrust::make_permutation_iterator(getStartPtr(qureg), indIter);
     auto probIter= thrust::make_transform_iterator(ampIter, functor_getAmpReal());
@@ -791,7 +791,7 @@ qreal thrust_statevec_calcProbOfMultiQubitOutcome_sub(Qureg qureg, vector<int> q
     auto indFunctor = functor_insertBits<NumQubits>(getPtr(sortedQubits), valueMask, qubits.size());
     auto probFunctor = functor_getAmpNorm();
 
-    auto rawIter = thrust::make_counting_iterator(0LL);
+    auto rawIter = thrust::make_counting_iterator(QINDEX_ZERO);
     auto indIter = thrust::make_transform_iterator(rawIter, indFunctor);
     auto ampIter = thrust::make_permutation_iterator(getStartPtr(qureg), indIter);
     auto probIter = thrust::make_transform_iterator(ampIter, probFunctor);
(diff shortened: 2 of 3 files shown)


otbrown:
@JPRichings Fix branch for your testing pleasure!

otbrown:
Single AMD GPU on ARCHER2 ✅:
```
otbz19@ln02:/work/z19/z19/otbz19/QuEST/QuEST> cat slurm-11307502.out

QuEST execution environment:
  precision:       2
  multithreaded:   1
  distributed:     1
  GPU-accelerated: 1
  GPU-sharing ok:  0
  cuQuantum:       0
  num nodes:       1

Testing configuration:
  test all deployments:  0
  num qubits in qureg:   6
  max num qubit perms:   25
  max num superop targs: 4
  num mixed-deploy reps: 10

Tested Qureg deployments:
  GPU + OMP + MPI

Randomness seeded to: 2726962016
===============================================================================
All tests passed (51879 assertions in 269 test cases)
```
and okay one failure on 4 GPUs, but I believe that's entirely unrelated to this change...
```
QuEST execution environment:
  precision:       2
  multithreaded:   1
  distributed:     1
  GPU-accelerated: 1
  GPU-sharing ok:  0
  cuQuantum:       0
  num nodes:       4

Testing configuration:
  test all deployments:  0
  num qubits in qureg:   6
  max num qubit perms:   25
  max num superop targs: 4
  num mixed-deploy reps: 10

Tested Qureg deployments:
  GPU + OMP + MPI

Randomness seeded to: 4273463917

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tests is a Catch2 v3.8.0 host application.
Run with -? for options

-------------------------------------------------------------------------------
rightapplyCompMatr
  validation
  targeted amps fit in node
-------------------------------------------------------------------------------
/work/z19/z19/otbz19/QuEST/QuEST/tests/unit/operations.cpp:1043
...............................................................................

/work/z19/z19/otbz19/QuEST/QuEST/tests/unit/operations.cpp:1083: FAILED:
  REQUIRE_THROWS_WITH( apiFunc(), ContainsSubstring("cannot simultaneously store") && ContainsSubstring("remote amplitudes") )
with expansion:
  "rightapplyCompMatr: Expected a density matrix Qureg but received a
  statevector." ( contains: "cannot simultaneously store" and contains: "remote
  amplitudes" )
with messages:
  minNumCtrls := 0
  numNewTargs := 5
  numQubits - minNumCtrls := 6
  ctrls := {  }
  targs := { 0, 1, 2, 3, 4 }

===============================================================================
test cases:   269 |   268 passed | 1 failed
assertions: 51315 | 51314 passed | 1 failed
```

I'll try a 31/32 qubit QFT too.

otbrown:
31-qubit QFT works and 32-qubit doesn't work on our AMD GPUs, as expected. I also checked and everything works fine with 32 qubits on 4 GPUs 👍

TysonRayJones:
Eep good catch! Thrust literal mischief had already [bitten me](https://github.com/QuEST-Kit/QuEST/blob/00ddd9397e408dd7b02771b6bc4a6d80b819be93/quest/src/gpu/gpu_thrust.cuh#L752-L758)!

To be extremely defensive, one could replace each `0LL` literal in the patch with a reference to e.g. `QINDEX_ZERO` defined somewhere like `bitwise.h`*, like is already done (albeit ) for the `1` literal: https://github.com/QuEST-Kit/QuEST/blob/00ddd9397e408dd7b02771b6bc4a6d80b819be93/quest/src/core/bitwise.hpp#L31
This would protect against future silent Thrust type issues if `qindex` was ever changed. And perhaps to evidence why that's a good idea, the `QINDEX_ONE` macro above is actually **wrong** since it treats `qindex` as _unsigned_, aha! You could correct that to `1LL` in this patch too if you fancied :^) 

Probably also worth then replacing that `#define` macro(s) with something explicitly typed to more securely avoid these literal-type issues, e.g.
```
// 0 remains agnostic to qindex type now
constexpr qindex QINDEX_ZERO = 0;
``` 

*It feels a little ill-fitting to define `QINDEX_ZERO` in `bitwise.hpp` rather than `types.h` or `precision.h` but the latter two are user-facing `¯\_(ツ)_/¯`

Let me know if you agree but don't have a sec, in which case I can add the changes to this PR!

TysonRayJones:
PS: I'll patch that "rightapplyCompMatr" error. It's because of this line...
https://github.com/QuEST-Kit/QuEST/blob/00ddd9397e408dd7b02771b6bc4a6d80b819be93/tests/unit/operations.cpp#L1043-L1047
which always uses a statevector to test the `"targeted amps fit in node"` validation, though the `rightapply*()` functions cannot accept statevectors, instead only density matrices. Because the `"was given a density matrix"` validation happens _before_ `"targeted amps fit in node"` validation, the latter intended triggered error was beaten out by the earlier unintended one.

TysonRayJones:
My changes are passing (caution: testing only the affected functions) with nvcc v12.8 🙏  Happy to squash and co-commit as you fancy!

otbrown:
Hi Tyson, 

Thanks for this! I like the principle of basing this on the `qindex` type, and was pondering similar over the weekend. Only note is that bitwise.hpp was probably using `1ULL` intentionally as bitwise shift is [undefined on some signed integers](https://en.cppreference.com/w/cpp/language/operator_arithmetic.html) until C++20. 

Having looked at bitwise.hpp I don't see anything that should be an issue, but we should keep it in mind if we get any _weird_ bugs 😆 

I'll retest this branch on our GPU systems and then merge, all being well!

otbrown:
Okay so there is still a rightapplyCompMatr test failure, but it is still unrelated to the GPU implementation -- it can be reproduced with any distributed build. I've spun that out into a new issue (#700) as all the _other_ tests pass on the AMD GPUs, and importantly I can run the 31/32 qubit jobs no problem. As long as @JPRichings tests on the CUDA platforms pass I think we can merge this safely 😄

JPRichings:
Updated branch passed tests on grace-hopper:

QuEST execution environment:
  precision:       2
  multithreaded:   1
  distributed:     0
  GPU-accelerated: 1
  GPU-sharing ok:  0
  cuQuantum:       0
  num nodes:       1

Testing configuration:
  test all deployments:  0
  num qubits in qureg:   6
  max num qubit perms:   25
  max num superop targs: 4
  num mixed-deploy reps: 10

Tested Qureg deployments:
  GPU + OMP

Randomness seeded to: 3270576438
===============================================================================
All tests passed (51882 assertions in 269 test cases)

otbrown:
Excellent, thanks all. I'll merge.

Issue #596, opened 2025-04-27
Improve the accuracy of Trotterisation with randomisation and commuting groups


TysonRayJones:
The function [`applyTrotterizedPauliStrSumGadget()`](https://github.com/QuEST-Kit/QuEST/blob/8678836cd29ac59688556399860aab712db1fe14/quest/src/api/operations.cpp#L1169) effects an approximation of `exp(i angle PauliStrSum)` using higher-order Suzuki-Trotter decomposition. Users can improve its accuracy by increasing the `order` and `reps` parameters.

The base accuracy however can be improved either/both by:
- randomising the effected order of the `PauliStrSum` terms within each repetition (as described [here](https://quantum-journal.org/papers/q-2019-09-02-182/)).
- grouping `PauliStrSum` terms into commuting groups which, when `exp(i term)` within are effected contiguously, induce no Trotter error.

Note that _optimally_ grouping commuting terms is [known](https://pubs.aip.org/aip/jcp/article-abstract/152/12/124114/954934/) to be NP-hard, but there exist a myriad of greedy/heuristic terms to find suboptimal groups. For example [this work](https://www.nature.com/articles/s41534-023-00697-6.pdf) greedily reduces the number of needed gadgets, offering both an accuracy and runtime improvement, and [this work](https://quantum-journal.org/papers/q-2020-09-12-322/).

The `applyTrotterizedPauliStrSumGadget()` function can be improved by:
- introducing randomisation, using the RNG in [`randomiser.cpp`](https://github.com/QuEST-Kit/QuEST/blob/main/quest/src/core/randomiser.cpp), which users can ergo control (to deterministically repeat a previous Trotter circuit) using routines like [`setSeeds()`](https://github.com/QuEST-Kit/QuEST/blob/53f8f3ad60e5b0171646ee8250c0e6ea65878b44/quest/src/api/debug.cpp#L37)
- pre-grouping `PauliStrSum` into (suboptimal) commuting groups. If sufficiently quick, this could be done each time the Trotter function is called. 
[...]

PR #702, opened 2025-11-12
Random permutation of strings in Trotterisation

vaferreiQMT:
Implements
 - random permutations of Pauli strings in a `PauliStrSum`
 - option to use random permutations during Trotterisation

See also #596

### DIFF
--- a/examples/extended/dynamics.c
+++ b/examples/extended/dynamics.c
@@ -153,11 +153,12 @@ int main() {
     int order = 4;
     int reps  = 5;
     int steps = 20;
+    bool randomise = true;
 
     for (int i=0; i<steps; i++) {
 
         // evolve qureg under (approx) exp(-i dt H)
-        applyTrotterizedUnitaryTimeEvolution(qureg, hamil, dt, order, reps);
+        applyTrotterizedUnitaryTimeEvolution(qureg, hamil, dt, order, reps, randomise);
 
         // calculate and report <O>
         qreal time = dt * (i+1);
@@ -187,7 +188,7 @@ int main() {
 
     // verify results by uninterrupted higher-order simulation to target time
     initPlusState(qureg);
-    applyTrotterizedUnitaryTimeEvolution(qureg, hamil, dt*steps, order+2, reps*steps);
+    applyTrotterizedUnitaryTimeEvolution(qureg, hamil, dt*steps, order+2, reps*steps, randomise);
     reportScalar("final <O>", calcExpecPauliStrSum(qureg, observ));
 
     // clean up
--- a/examples/extended/dynamics.cpp
+++ b/examples/extended/dynamics.cpp
@@ -150,11 +150,12 @@ int main() {
     int order = 4;
     int reps  = 5;
     int steps = 20;
+    bool randomise = true;
 
     for (int i=0; i<steps; i++) {
 
         // evolve qureg under (approx) exp(-i dt H)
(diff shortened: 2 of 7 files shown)


otbrown:
Hi Vasco, 

Thanks for this -- just wanted to confirm that I've seen it and will try to review once I'm back from SC (which is next week!).

Cheers,
Oliver

TysonRayJones:
Hi Vasco and Oliver,

Here are some thoughts before merging:

- [ ] Making a new `randomiseString` parameter constitutes an API break - but I personally think that's fine and doesn't warrant a new major version.
- [ ] The new `randomiseString` parameter needs to be added to the docstrings.
- [ ] Is `randomiseString` the most illustrative name? I worry the `String` suffix doesn't clarify much - the code is shuffling the order of the prescribed Pauli rotations within each Trotter repetition - so I would personally advocate for simplifying it to `randomise`.
- [ ] The current implementation mutates the passed `PauliStrSum`, so a user printing their `PauliStrSum` will see it changed pre- and post-call of any of the Trotter functions. If that's a desired/acceptable postcondition, we need to doc that. If that's undesirable, it'd be straightforward to additionally maintain a list of the original indices with respect to the current mutation and restore the original order before function return. That requires `O(#terms)` temporary memory which may be justifiable (e.g. the Lindblad functions use greater temp mem).

otbrown:
Hi both,

- On the API break, I agree, I think this is sufficiently new functionality in QuEST that we can justify being a bit loose with the semantic versioning.
- Rather than `randomise` or `randomiseString` I would propose something more verbose, like `randomisePauliOrder`, though this is not a hill I care to die on :)
- To Tyson's point about printing the sum, I'm a bit torn -- I can see the sense in not changing something the user has defined themselves, but I _also_ don't like the idea of not giving them the opportunity to inspect what QuEST really ran. As such my vote is for documenting that this function mutates the `PauliStrSum`, and leaving it mutated so the user can inspect what happened.
- Only additional comment is that there should be at least some basic testing, as discussed on the call the other day Vasco!

vaferreiQMT:
Thank you both for the feedback. I should have some time to implement your suggestions and update the PR next week or so.

Cheers

vaferreiQMT:
Howdy, I've finally had some time to update the PR.

A couple notes:
- I've added a few lines on the Trotter docs about randomisation, including a warning about `sum` being mutated
- I updated the name to `permutePaulis` which I think should hopefully be more self-explanatory whilst also ensuring there isn't any confusion between "order" (as in the degree of the Trotter expansion) and the "order" (as in permutation of Pauli strings).
- I've added a very rough test, which hopefully can be integrated more neatly once the other Trotterisation tests are implemented. For now it serves as a quick sanity-check that randomisation is in fact improving the Trotterisation when used in the regime which we expect it to behave well.

Happy to also make any additional tweaks or alterations if anything jumps out at you.

Also a heads up, I am working on improved grouping and ordering of Trotter terms, though I still have to tidy that up a bit and will open a draft PR once it is more complete.

otbrown:
Hi @vaferreiQMT,

This looks good to me! I see there were some test failures which _look_ like issues with the runner not the code. After trying and failing to find a way to rerun those checks I checked it manually using the ARCHER2 AMD GPU testbed. As a bonus I can confirm that tests pass there too!

I'll look at merging #706 next, which will add some tests for the underlying evolution functions.

Issue #604, opened 2025-04-30
Optimise `applyQFT()`


TysonRayJones:
The current implementation of [`applyQuantumFourierTransform()`](https://github.com/QuEST-Kit/QuEST/blob/53f8f3ad60e5b0171646ee8250c0e6ea65878b44/quest/src/api/operations.cpp#L1918) merely effects the canonical gates of the QFT in-turn. An optimised routine is possible, whereby we merge contiguous phase gates into a single, diagonal operator.  In QuEST `v3`, this was [implemented](https://github.com/QuEST-Kit/QuEST/blob/eb74b521b958a9ece945032571f5ef5e9232a16f/QuEST/src/QuEST_common.c#L830) using the (now deprecated and absolutely awful) [`applyPhaseFunc()`](https://github.com/QuEST-Kit/QuEST/blob/eb74b521b958a9ece945032571f5ef5e9232a16f/QuEST/src/CPU/QuEST_cpu.c#L4406) routine.

In QuEST `v4`, an equivalent diagonal operator _could_ be effected by a [`applyFullStateDiagMatr`](https://github.com/QuEST-Kit/QuEST/blob/53f8f3ad60e5b0171646ee8250c0e6ea65878b44/quest/src/api/operations.cpp#L556) using a temporarily allocated `applyFullStateDiagMatr`. However, this incurs gratuitous memory allocation costs given that the elements are compile-time known.

We should instead implement the necessary diagonal operator as a bespoke routine. This requires re-deriving it (since removed last-minute from the [distributed manuscript](https://arxiv.org/abs/2311.01512)) and updating `applyQuantumFourierTransform()` to make use of it. The algorithm is very cute, compatible with both distribution, and density matrices, and non-contiguous arbitrarily-ordered target qubits. As such, it may be a superior solution to performing the FFT upon the constituent amplitudes directly.

> This may serve as an excellent project for a student and worthy of publication 🙌

TysonRayJones:
Worth pointing out that QFT on a non-contiguous, arbitrarily-ordered subset of qubits _can_ make use of `FFT`; we simply swap active qubits to the lowest ordered positions, apply many instances of FFT in parallel upon all `2^#targets` sub-arrays, and SWAP back. This works also on density matrices whereby we process each column independently. However...
- this precludes full-state QFT in distributed settings; we could not target `log2(#nodes)` qubits.
[...]

PR #705, opened 2026-01-06
Add inverse to QFT implementation

vaferreiQMT:
Howdy, I hope your holidays and the start of the new year have been great!

While the refactor of the QFT implementation in #604 remains in the backlog, I think it would still prove useful to have an inverse QFT for completeness. Since I needed this for another project and it is a fairly minor change I figured it might be worth tidying it and submitting a PR.

Main changes:
- **Inverse QFT**: adapted from the forward implementation. Controlled via an `inverse` parameter, which is a breaking change, but I think it makes sense to keep the QFT and inverse together. Happy to split it if you disagree.
- **Inverse QFT unit tests**: using the existing tests for the QFT with an additional generator for the inverse parameter.
- **QFT docs**: documentation for the QFT methods. Maybe not to full fit and finish, but hopefully better than nothing.

Cheers!

### DIFF
--- a/quest/include/operations.h
+++ b/quest/include/operations.h
@@ -2416,8 +2416,8 @@ extern "C" {
  * 
  * @formulae
  * 
- * Let @f$ N @f$ = @p numTargets, the @f$ N @f$ qubit Quantum Fourier Transform maps each
- * computational basis state belonging to the targeted qubits, @f$ \ket{j} @f$, according to
+ * Letting @f$ N @f$ = @p numTargets, the @f$ N @f$ qubit Quantum Fourier Transform maps each
+ * computational basis state of the targeted qubits, @f$ \ket{j} @f$, according to
  * @f[ 
         \ket{j} \rightarrow \frac{1}{\sqrt{2^N}} \sum_{k=0}^{2^N-1} e^{2 \pi i j k / 2^N} \ket{k}.
  * @f]
@@ -2433,6 +2433,7 @@ extern "C" {
  * @throws @validationerror
  * - if @p qureg is uninitialised.
 *  - if @p targets are invalid qubit indices.
+*  - if @p targets are not unique.
  * - if @p numTargets < 1.
  * @see
  * - applyFullQuantumFourierTransform()
--- a/quest/include/operations.h
+++ b/quest/include/operations.h
@@ -25,6 +25,8 @@
 #include "quest/include/matrices.h"
 #include "quest/include/channels.h"
 
+#include <stdbool.h>
+
 #ifdef __cplusplus
     #include <vector>
 #endif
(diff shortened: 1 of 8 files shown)


quest/include/operations.h
```diff
@@ -2406,14 +2408,70 @@ extern "C" {
...
-void applyQuantumFourierTransform(Qureg qureg, int* targets, int numTargets);
+/** @notyetdoced
+ * @notyetvalidated
+ *
+ * Applies the Quantum Fourier Transform to the specified @p targets of @p qureg.
+ * Alternatively, applies the Inverse Quantum Fourier Transform according to @p inverse.
+ * 
+ * @formulae
+ * 
+ * Let @f$ N @f$ = @p numTargets, the @f$ N @f$ qubit Quantum Fourier Transform maps each
```
TysonRayJones: Let -> Letting

quest/include/operations.h
```diff
@@ -2406,14 +2408,70 @@ extern "C" {
...
+ * computational basis state belonging to the targeted qubits, @f$ \ket{j} @f$, according to
```
TysonRayJones: belonging to -> of

quest/include/operations.h
```diff
@@ -2406,14 +2408,70 @@ extern "C" {
...
+        \ket{j} \rightarrow \frac{1}{\sqrt{2^N}} \sum_{k=0}^{2^N-1} e^{-2 \pi i j k / 2^N} \ket{k}.
+ * @f]
+ *
+ * @param[in,out] qureg      the state to modify.
+ * @param[in]     targets    the indices of the target qubits.
+ * @param[in]     numTargets the length of list @p targets
+ * @param[in]     inverse    whether to apply the inverse QFT or forward QFT
+ * @throws @validationerror
+ * - if @p qureg is uninitialised.
+*  - if @p targets are invalid qubit indices.
```
TysonRayJones, line 2425: add:
```
if @p targets are not unique.
```

quest/include/operations.h
```diff
@@ -2406,14 +2408,70 @@ extern "C" {
...
+ * - if @p numTargets < 1.
+ * @see
+ * - applyFullQuantumFourierTransform()
+ * @author Vasco Ferreira
+ */
+void applyQuantumFourierTransform(Qureg qureg, int* targets, int numTargets, bool inverse);
```
TysonRayJones, line 2432: Technically, type `bool` is only gauranteed to be defined in `C` when including the `<stdbool.h>` header. That header is in the standard since C99 however. @otbrown thoughts on this? This is the first `bool` in the QuEST API (ignoring measurement `outcomes` passed as `int`). Should we make `inverse` an `int` and validate it to be `==0` or `==1` (like measurement outcomes), or just include the header?

vaferreiQMT: Happy to update this, just a note that PR  #702 also included a `bool` (and the `<stdbool.h>` header for C), so if we do end up going with the `int` and validation route, it might make sense to update that too.

quest/include/operations.h
vaferreiQMT: Updated in 3b07c0c.

quest/include/operations.h
vaferreiQMT: Updated in 3b07c0c.

quest/include/operations.h
vaferreiQMT: Updated in 3b07c0c.

quest/include/operations.h
otbrown: Good catch @TysonRayJones! 

My thoughts are that:
- We are already using the C11 standard, so introducing `stdbool.h` doesn't introduce any build issues or push our requirements.
- Practically, bool vs int makes very little difference as they are somewhat interchangeable in C, _however_ bool has a semantic benefit, as it better expresses _intent_.

So in my opinion we should include stdbool.h and use bool here.

vaferreiQMT:
Hiya, just checking in on this PR. I think all the comments above should have been resolved.

quest/include/operations.h
```diff
@@ -2406,14 +2408,71 @@ extern "C" {
...
+        \ket{j} \rightarrow \frac{1}{\sqrt{2^N}} \sum_{k=0}^{2^N-1} e^{2 \pi i j k / 2^N} \ket{k}.
+ * @f]
+ * Similarly the Inverse Quantum Fourier Transform maps each basis state like
+ * @f[ 
+        \ket{j} \rightarrow \frac{1}{\sqrt{2^N}} \sum_{k=0}^{2^N-1} e^{-2 \pi i j k / 2^N} \ket{k}.
+ * @f]
+ *
+ * @equivalences
+ *
+ * - This function wraps applyQuantumFourierTransform() with all the qubits in the @p qureg as @p targets.
```
TysonRayJones: Note `@p targets` here causes doxygen to throw a warning, because `targets` is not a parameter of `applyFullQuantumFourierTransform`.

TysonRayJones: (fixed in conflict fix)

TysonRayJones:
Nice one! 🎉

vaferreiQMT:
Howdy, I've just noticed that the final merge of `devel` into this branch missed some of the changes to the new validation tests added in #706, which is causing the build to fail. I will make a PR for a fix (just adding the `inverse` to the QFT calls in the validation tests).

EDIT: fixes in PR #710

TysonRayJones:
Oops, possibly my fault - the never ending joys of git! 🎉

PR #710, opened 2026-02-19
fix: add missing inverse argument to QFT validation tests

vaferreiQMT:
Adds missing `inverse` argument to the validation tests for the QFT, which were causing build to fail.


TysonRayJones:
I see compilaton has passed, which was the only issue - I will recklessly merge this in (I like to live dangerously 😎 )

Issue #704, opened 2025-12-04
HIP loses the imaginary part on ROCM 7


eessmann:
When building and running QuEST with ROCM 7, we lose the imaginary components of the state vector.  For example, here’s a qureg initialised with a random pure state vector and we still lose the imaginary components regardless of how I initialise it.

```bash
Qureg (4 qubit statevector, 16 qcomps, 360 bytes in gpu):
    0.23598     |0⟩
    0.19536     |1⟩
    -0.25405    |2⟩
    -0.14684    |3⟩
    0.11705     |4⟩
    -0.045575   |5⟩
    -0.0095484  |6⟩
    0.42606     |7⟩
    0.082763    |8⟩
    -0.10199    |9⟩
    -0.19393    |10⟩
    0.027218    |11⟩
    -0.019831   |12⟩
    -0.21442    |13⟩
    -0.11921    |14⟩
    -0.090414   |15⟩

Total probability: 0.5017
```

eessmann:
There seem to be two ways to handle complex numbers in CUDA:
from `<cuda/std/complex>`
```c++
cuda::std::complex<double> a
```
Basically a works like the c++ complex number and has all the nice overloads and documentation.
from `cuComplex.h`
```c++
cuComplex c
```
I can't find any real documentation on it, apart from that it exists.

TysonRayJones:
Oh that's shocking! Are the unit tests failing in that configuration? I believe they _should_ because they compare to a CPU implementation which (often) has a hardcoded non-zero imaginary component (e.g. [here](https://github.com/QuEST-Kit/QuEST/blob/9d7618d7263e3bfba433b88cf1eac0647f08fa0a/tests/utils/qvector.cpp#L140)), but just making sure.

I lean away from `cuda::std::complex<double>`, given its non-standardness (I'd sooner choose [`thrust::complex`](https://nvidia.github.io/cccl/thrust/api/structthrust_1_1complex.html)). But that's moot because the current `cuComplex` _is_ the recommended way to go (despite the unusual doc situation). HIP [explicitly supports it](https://rocm.docs.amd.com/projects/HIPIFY/en/docs-5.2.3/tables/cuComplex_API_supported_by_HIP.html), and it's what the NVIDIA cuQuantum software suites primarily use (see example [here](https://docs.nvidia.com/cuda/cuquantum/22.05.0/custatevec/getting_started.html)). So it's certainly too early to go seek another complex type hehe.

[...]

PR #718, opened 2026-03-16
Fix HIP install failures in CI

otbrown:
Idea for fix brazenly stolen from here: https://github.com/actions/runner/issues/3720

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -17,8 +17,8 @@ on:
       - devel
   pull_request:
     branches:
-      - main
-      - devel
+      #      - main
+      #      - devel
 
 
 jobs:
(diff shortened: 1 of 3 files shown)


otbrown:
...damn it. On the plus side, it's a new error!

otbrown:
Hi @TysonRayJones -- that's got it I think. We're now using the current recommended installation instructions from [AMD](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/install/install-methods/package-manager/package-manager-ubuntu.html). 

Note that although it compiles, tests would certainly fail due to #704!

TysonRayJones:
Wew brilliant! Oh to see that beautiful green CI tick 🤩 Happy for it to be squash merged!

PR #724, opened 2026-04-15
Enabling GPU direct communication for Cray-MPICH

JPRichings:
This needs testing but here is an initial suggestion for checking support for GPU enabled MPI in CRAY_MPICH this should enable direct inter GPU comms for a wider range of systems in the HPC space that don't always support OpenMPI for there scale out network.

Plan to test on MI210 nodes on ARCHER2 when they are available.

### DIFF
--- a/quest/include/environment.h
+++ b/quest/include/environment.h
@@ -36,7 +36,7 @@ typedef struct {
     int isMultithreaded;
     int isGpuAccelerated;
     int isDistributed;
-    int isMPIGPUAware;
+    int isMpiGpuAware;
 
     // deployment modes which cannot be directly changed after compilation
     int isCuQuantumEnabled;
--- a/quest/include/environment.h
+++ b/quest/include/environment.h
@@ -36,6 +36,7 @@ typedef struct {
     int isMultithreaded;
     int isGpuAccelerated;
     int isDistributed;
+    int isMPIGPUAware;
 
     // deployment modes which cannot be directly changed after compilation
     int isCuQuantumEnabled;
--- a/quest/src/api/environment.cpp
+++ b/quest/src/api/environment.cpp
@@ -147,6 +147,7 @@ void validateAndInitCustomQuESTEnv(int useDistrib, int useGpuAccel, int useMulti
     globalEnvPtr->isDistributed       = useDistrib;
     globalEnvPtr->isCuQuantumEnabled  = useCuQuantum;
     globalEnvPtr->isGpuSharingEnabled = permitGpuSharing;
+    globalEnvPtr->isMPIGPUAware = comm_set_isMpiGpuAware();
 
     // bind distributed info
     globalEnvPtr->rank     = (useDistrib)? comm_getRank()     : 0;
--- a/quest/src/comm/comm_config.cpp
+++ b/quest/src/comm/comm_config.cpp
@@ -12,6 +12,7 @@
  * @author Tyson Jones
  */
 
+#include "quest/include/environment.h"
 #include "quest/include/config.h"
 #include "quest/include/types.h"
 
@@ -19,6 +20,7 @@
 #include "quest/src/core/errors.hpp"
 
 #include <cstdlib>
+#include <string>
 
 #if COMPILE_MPI
     #include <mpi.h>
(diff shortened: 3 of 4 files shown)


JPRichings:
Should have updated this before:

Hardware: 1 node 4 gpus Mi210 ARCHER2

before (no MPI direct comms):

QFT run time: 9.49068s
Total run time: 10.2344s

after (with MPI direct comms):

QFT run time: 2.29135s
Total run time: 2.92993s

I think it might be worth switching on GPU direct comms... ;)

JPRichings:
Test all failing due to:
```
[ 80%] Building CXX object _deps/catch2-build/src/CMakeFiles/Catch2.dir/catch2/matchers/catch_matchers_predicate.cpp.o
/Users/runner/work/QuEST/QuEST/quest/src/comm/comm_config.cpp:75:5: error: use of undeclared identifier 'MPI_Get_library_version'
   75 |     MPI_Get_library_version(version_string, resultlen);
      |     ^
1 error generated.
```

- [x] I will look to add macro guard around this mpi call to fix non-mpi compilation.
- [x] need to set some sensible default buffer sizes
- [x] Need to remove the debugging messages for the code

JPRichings:
need to return NONE for comm_whichMpi and false for comm_set_isMpiGpuAware

quest/include/environment.h
```diff
@@ -36,6 +36,7 @@ typedef struct {
     int isMultithreaded;
     int isGpuAccelerated;
     int isDistributed;
+    int isMPIGPUAware;
```
TysonRayJones: Maybe this should be `isMpiGpuAware` to be consistent with `isGpuSharingEnabled`, `comm_set_isMpiGpuAware`, etc

otbrown: Resolved in 3550cd5

TysonRayJones:
Hmm I think this diff is unnecessarily convoluted. It seeks to extend the systems for which we can detect GPU-awareness, but creates a `comm_config.cpp` dependency on an API env function:
```cpp
bool comm_isMpiGpuAware() {
	return getQuESTEnv().isMPIGPUAware;
}
```
and performs brittle string processing to establish the MPI version...
```cpp
enum Mpi_version {NONE, OPENMPI, CRAYMPICH};

int comm_whichMpi() {

    #ifdef COMPILE_MPI
        char version_string[MPI_MAX_LIBRARY_VERSION_STRING];
    #else
        char version_string[];
    #endif

    int resultlen[] = {0};

    #ifdef COMPILE_MPI
        MPI_Get_library_version(version_string, resultlen);
    #endif
    enum Mpi_version version = NONE;

    // Check if Openmpi used
    #ifdef OPEN_MPI
        version = OPENMPI;
    #endif

    // Check if Cray MPI used 
	
    const char* cray_string = "CRAY MPICH";

    std::string v_string = version_string;

    if (v_string.find(cray_string) != string::npos) {
        version = CRAYMPICH;
    }

    return version;

}
```
just so that it can merely look for a single env variable, `MPICH_GPU_SUPPORTED_ENABLED`:
```cpp
bool comm_set_isMpiGpuAware() {

   int mpi_lib = comm_whichMpi();
  
    if(OPENMPI==mpi_lib) {
        // definitely not GPU-aware if compiler declares it is not
        #if defined(MPIX_CUDA_AWARE_SUPPORT) && ! MPIX_CUDA_AWARE_SUPPORT
            return false;
        #endif

        // check CUDA-awareness at run-time if we know it's principally supported
        #if defined(MPIX_CUDA_AWARE_SUPPORT)
            return (bool) MPIX_Query_cuda_support();
	#endif
    }

    if(CRAYMPICH==mpi_lib) { 
          
       const char* var = std::getenv("MPICH_GPU_SUPPORT_ENABLED");
       
       return (bool) var;
    }

    // if we can't ascertain CUDA-awareness, just assume no to avoid seg-fault
    return false;
}
```

In my understanding, this is all unnecessary because the check of "is GPU aware" is always _**fail safe**_. If we _cannot_ detect an affirmative "yes", then we default to "no". Checking the MPI versions is redundant, given we can just check for the macros directly - when they are _not_ defined, the check is null (_NOT_ false) and we safely proceed.

The original `comm_isMpiGpuAware` function was:
```cpp
bool comm_isMpiGpuAware() {

    // definitely not GPU-aware if compiler declares it is not
    #if defined(MPIX_CUDA_AWARE_SUPPORT) && ! MPIX_CUDA_AWARE_SUPPORT
        return false;
    #endif

    // check CUDA-awareness at run-time if we know it's principally supported
    #if defined(MPIX_CUDA_AWARE_SUPPORT)
        return (bool) MPIX_Query_cuda_support();
    #endif

    // if we can't ascertain CUDA-awareness, just assume no to avoid seg-fault
    return false;
}
```
If my understanding is correct, to support CRAY MPICH, we just need to introduce 3 lines before the `return false`, like:
```cpp
    // check whether an MPICH env-var indicates support (we assume it never lies!)
    static const char* var = std::getenv("MPICH_GPU_SUPPORT_ENABLED"); // load once
    if (var && std::atoi(var))
        return true;
```

The only _functional_ difference between that and this PR's diff is that the PR will _never try_ to find the MPICH env-var when it knows OpenMPI is used. That doesn't add any robustness (the OpenMPI checks were safe, since undefined macros in C default to falsiness), _except_ if we believe OpenMPI systems might have an erroneous `MPICH_GPU_SUPPORT_ENABLED=1` env-var. And of course that is **_absurd_**, and much _less likely_ than the _true_ danger inherent to both these solutions; that an MPICH user will have somehow erroneously overriden `MPICH_GPU_SUPPORT_ENABLED=1`. 

(Note too this branch doesn't compile, beyond just the `isMPIGPUAware` typo; `MPI_MAX_LIBRARY_VERSION_STRING` is not defined in Clang's MPI)

It seems best to close the PR and create a fresh one with the 3 lines mentioned above. Have I understood all this right?

TysonRayJones:
Note I like the idea of attaching `isMpiGpuAware` to the `environment` at runtime. We should keep that, but then of course report it within `reportQuESTEnv()` and `getEnvironmentString()` so users can see it! I can add all that in a new PR

(Also note to self; rename `getEnvironmentString()` to `getQuESTEnvString()` - iirc this function exists for a pyQuEST hook, and @rrmeister can tolerate an API break ;) )

otbrown:
Managed to catch @JPRichings this morning to check there wasn't an underlying reason that we needed the version, and he confirmed that it was only to follow the pattern of the existing OpenMPI implementation. He also likes your version more, so if you're happy, we're happy :)

> Note too this branch doesn't compile, beyond just the isMPIGPUAware typo; MPI_MAX_LIBRARY_VERSION_STRING is not defined in Clang's MPI

I'm guessing that was Intel MPI? As it seems to be defined in both OpenMPI and MPICH. Doesn't matter since we're not using it any more, but good to know for the future!

I'll close!

otbrown:
Superseded by #766.

PR #728, opened 2026-04-17
Trotterisation test update

### LINKED ISSUE #715: Fix Trotter unit tests
The current Trotter function unit tests call `REQUIRE` inside loops with $2^{16}$ iterations, inflating the assertion count. As a result, a serial-deployment `"applyTrotterizedImaginaryTimeEvolution"` test results in `(262153 assertions in 1 test case)`. 

Relatedly, they appear to be comparing two `Qureg`s per-amplitude, via
```
// iterating over i
qcomp other_amp = getQuregAmp(qureg, i);
qreal other_mag = other_amp.real() * other_amp.real() +  other_amp.imag() * other_amp.imag();
```
(instead of simply `calcProbOfBasisState`), but it is unclear why the `Qureg` must be compared via magnitude instead of the existing `Qureg`-wide amplitude comparisons. Fix this!

otbrown:
Resolves #715

### DIFF
--- a/.github/workflows/test_paid.yml
+++ b/.github/workflows/test_paid.yml
@@ -89,7 +89,7 @@ jobs:
       # don't test Trotter functions (ctest is func-name regex, catch is tag)
       test_exec: ./tests/tests
       ctest_flag: "-E Trotter"
-      catch_flag: "~[trotterisation]"
+      catch_flag: '"~[trotterisation]"'
 
       # GPU runner has a Tesla T4 16 GB
       cuda_arch: 75
--- a/.github/workflows/test_paid.yml
+++ b/.github/workflows/test_paid.yml
@@ -1,6 +1,13 @@
 # Tests execution of v4 unit and integration tests
 # on paid runners using multithreading, GPU-
 # acceleration and distribution, only on Linux. 
+#
+# Note that the blow tests are deliberately
+# excluding Trotter functions because they are
+# slow and expensive, and the functions merely
+# invoke other functions existed elsewhere, in
+# a backend agnostic way.
+#
 # As of 12/03/2025, these "large runners" cost:
 # - 16c/min for each ARM runner (OMP, MPI)
 # - 7c/min for each GPU runner (CUDA, cuQuantum)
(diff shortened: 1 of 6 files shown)


otbrown:
Bloody quad precision. I suspect calculating exponents is the underlying culprit here. We're probably checking against an overly optimistic absolute error in these. Will tweak it up for these tests.

otbrown:
Ah, I see `REQUIRE_AGREE` uses tolerances hard-coded in tests/utils/compare.cpp, not validation epsilon... Also it's _only_ the serial and quad precision tests that fail. What a confounding edge case 😑

otbrown:
Okay, when tested locally all quad precision configurations failed, so I've applied the epsilon increase for all deployments, not just single-threaded.

otbrown:
Hi @TysonRayJones, hopefully this version of the Trotter tests sparks joy!

We're now using a bit more of the general QuEST testing machinery, and testing across deployments. The unitary time evo tests are a bit trickier, as I really can't think of a way to do that that doesn't involve fixed numbers of qubits, but again I've updated them to work across deployments and test on density matrices as well as statevectors.

TysonRayJones:
Wew amazing! 🎉 🎉  I'll do a micro change to the var-names in `cache` and `compare` for src consistency (no need for coauth ofc).

Note in quad-precision and with multithreading on my machine, I get a failure:
```
compare.cpp:177: FAILED:
explicitly with messages:
  scalar := (5.36958,0)
  reference := (5.36958,0)
  absolute_difference := 5.43153e-10
  absolute_epsilon := 1e-10
  relative_difference := 1.01154e-10
  relative_epsilon := 1e-10
```
Happy to just bump both epsilons by `*10` ¯\\_(ツ)_/¯

Finally, these tests are (relatively) slow, and we may wish to avoid them in paid CI runners (if they hypothetically return) since the trotter functions just call separately tested backend functions; they are ergo "backend agnostic". So we don't actually _have_ to run these tests cross-platform to know they'll pass (although I'd so for rigour if time was free, of course). I'll add a flag to the github workflows to skip trotter tests in the paid CI

otbrown:
All sounds good to me, thanks Tyson!

TysonRayJones:
@otbrown this is ready to merge if you're happy with my changes above

otbrown:
Looks good to me! Thanks Tyson.

PR #736, opened 2026-04-24
Set gpu tpb

### LINKED ISSUE #735: Tunable GPU block sizes
Noticed that 

https://github.com/QuEST-Kit/QuEST/blob/9d7618d7263e3bfba433b88cf1eac0647f08fa0a/quest/src/gpu/gpu_kernels.cuh#L50

is fixed for all target hardware and is a bit large for common tuning recommendations.

Plan to change this to allow a compile time default and a setter-getter interface to allow performance tuning tests.

otbrown:
Creating a facility for users to runtime set threads per block for tuning the GPU implementation. NOTE: only applies to kernels that are not handled by Thrust, which does its own thing. Resolves #735. 

I considered and rejected the idea of creating a symmetric interface for the CPU for users who don't know `OMP_NUM_THREADS` or `omp_set_num_threads()` exist, but that's much riskier as the point of truth is external (in the OpenMP runtime).

TODO:
- [ ] ~~Should gpu_getNumThreadsPerBlock return a `qindex`? Probably~~.
- [ ] ~~Create a new home for user facing API, as environment doesn't really make sense.~~ Kicking this back to next release.
- [x] Add a compile time default value -- that way expert maintainers can compile a tuned default into a library which is used on a system.
- [ ] ~~Query seemingly unused branch at https://github.com/QuEST-Kit/QuEST/blob/b7d4a291927ba64d24bcac265ad7618a8ee218e0/quest/src/gpu/gpu_subroutines.cpp#L453~~
- [x] Add TPB to QuEST GPU environment reporting.
- [x] Add tests for new interface.
- [x] @JPRichings To check if this is really worthwhile, but please wait a week to tell me if it isn't.
  -  It is!


otbrown:
Rudimentary testing done with:
```C++
#include <cstdio>
#include "quest.h"

int main (void)
{
  const int NQUBITS = 24;
  const int TPB = 32;


  initQuESTEnv();
  reportQuESTEnv();

  std::printf("Initial number of threads per block: %d\n", getQuESTGpuThreadsPerBlock());

  setQuESTGpuThreadsPerBlock(TPB);
  std::printf("New number of threads per block: %d\n", getQuESTGpuThreadsPerBlock());

  Qureg qureg = createForcedQureg(NQUBITS);

  std::printf("Initialising Qureg.\n");
  initPlusState(qureg);
  reportQureg(qureg);

  std::printf("Applying Quantum Fourier Transform.\n");
  applyFullQuantumFourierTransform(qureg, false);
  reportQureg(qureg);

  destroyQureg(qureg);
  finalizeQuESTEnv();

  return 0;
}
```

JPRichings:
Why would  `gpu_getNumThreadsPerBlock` be a qindex this is not a quantum quantity. `uint` should be fine ( I am sure there is a recommendation from the cuda api we can match.

TysonRayJones:
Is there an advantage to users having to set this as a runtime hyperparameter? My (mostly undeveloped) belief is we can use [occupancy tools](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__OCCUPANCY.html) (alluded to [here](https://github.com/QuEST-Kit/QuEST/blob/b7d4a291927ba64d24bcac265ad7618a8ee218e0/quest/src/gpu/gpu_kernels.cuh#L50-L72)) to automate this. I definitely shy from giving users a greater onus to optimise for their settings (like other prolific softwares), which the v4 overhaul was supposed to avoid (via e.g. the autodeployer).

Note too that the kernels so far are very primitive - each thread handles the updating of the minimum possible number of amplitudes (often just _one_!). I quite like that because it's very readable and simple (great for an open-source scientific project) but _is_ an obvious site for optimisation.

> Why would gpu_getNumThreadsPerBlock be a qindex this is not a quantum quantity. uint should be fine ( I am sure there is a recommendation from the cuda api we can match.

It's true that it will never be anywhere as big as the quantities `qindex` is expected to store (like the number of basis states), but I have already used it in places where I thought an `int` might be insufficient. Inoffensive either way as `uint` or `qindex` imo

JPRichings:
Hi Tyson,

I just noticed the fixed value to 128 and have a feeling that it was large.

I just wanted a handle so I could write a benchmark so we can easily automate performance tuning ourselves.

I have not played with the occupancy tools but I should take a proper look as this might solve this automatically.

My other concern is that there are differences between Nvidia and AMD on optimal sizes due to hardware differences so we might not be able to reply on the occupancy tuning in all cases unless this becomes available on all platforms.

quest/src/cpu/cpu_config.cpp
```diff
@@ -79,9 +79,7 @@ int cpu_getAvailableNumThreads() {
 #if COMPILE_OPENMP
     int n = -1;
 
-    #pragma omp parallel shared(n)
-    #pragma omp single
-    n = omp_get_num_threads();
+    n = omp_get_max_threads();
```
TysonRayJones, line 82: Isn't this functionally wrong? We wish to return the number of available threads as set by the user, and which is the default adopted by our openmp pragmas. If you call `omp_get_max_threads()` outside a parallel region, won't it just return `1`?

quest/src/gpu/gpu_subroutines.cpp
```diff
@@ -141,7 +140,8 @@ qindex gpu_statevec_packAmpsIntoBuffer(Qureg qureg, vector<int> qubits, vector<i
 #if COMPILE_CUDA || COMPILE_CUQUANTUM
 
     qindex numThreads = qureg.numAmpsPerNode / powerOf2(qubits.size());
-    qindex numBlocks = getNumBlocks(numThreads);
+    const int NUM_THREADS_PER_BLOCK = gpu_getNumThreadsPerBlock();
```
TysonRayJones: If we opt for this, why is `NUM_THREADS_PER_BLOCK` capitalised like a constant? It's runtime

quest/src/cpu/cpu_config.cpp
JPRichings: From an EPCC colleague:

```
omp_set_num_threads() set the value of the nthreads internal control variable, but omp_get_num_threads() does not get this value ( but omp_get_max_threads() does).
```

Standards aren't immune to issues.

quest/src/gpu/gpu_subroutines.cpp
JPRichings: I agree capitalisation here bad.

TysonRayJones:
> I just noticed the fixed value to 128 and have a feeling that it was large.

I guess it's very GPU specific! I think `128` was motivated by thinking of `CC=3`, which has a max active blocks per SM of `16`, and a max active threads per SM of `2048`. So using `128` threads per block perfectly maximally occupies the SMs (when there are enough amplitudes to admit more than 16 blocks per SM, of course!) 

For illustration, the next smallest size is `96` (it must be a multiple of `32`, else threads within a warp will be idle), which yields a number of active threads of `16 * 96 = 1536`, which wastes `2048 - 1536 = 512` threads per SM!

Of course, newer GPUs support more active blocks per SM (even when the max active threads per SM is unchanged). E.g. CC=8 supports up to 32 active blocks per SM, so we could shrink to `64` threads per block while achieving the same occupancy - but I don't have a great intuition for the effect when we're memory-bandwidth bound.

Certainly seems prudent to consult a CUDA runtime API, if that doesn't hurt our AMD compatibility!

quest/src/cpu/cpu_config.cpp
otbrown: Yeah as James indicates `omp_get_num_threads` outside a parallel region will return 1, but `omp_get_max_threads` returns `OMP_NUM_THREADS` _or_ whatever was last set using `omp_set_num_threads`. Parallel regions without a `num_threads` clause then use that value.

quest/src/gpu/gpu_subroutines.cpp
otbrown: It's const in scope 😉 apologies, accidentally following my own style guide there rather than the QuEST one. I'll `%s/NUM_THREADS_PER_BLOCK/numThreadsPerBlock/g` it.

otbrown:
Apologies, _probably_ won't get to look at this again this week, but very happy to set this value programmatically if it can be done! 

As it's architecture dependent, we definitely do need a way to adjust it, and ideally both at runtime and compile time. At compile time, so kindly HPC support teams can compile and maintain a tuned version, and at runtime, so they can scan through values without having to recompile in between. I'll have a chat with James abour approaches later this week!

I 100% agree that we don't really want unknowing users messing around with this. I think something like an `architecture.h` or `perftune.h` or similar might be the best solution. A set of functionality that we explicitly document is for _users who know what they are doing_ to tune the performance of the library for a specific architecture. It might be this is the only value in there for the time being, but for slingshot-11 _reasons_ we need to add a parameter capping the total in-flight data and this would be a good spot for that too.

quest/src/cpu/cpu_config.cpp
TysonRayJones: Oh nice, my brain hadn't even noticed the change of `num` to max` 😅

TysonRayJones:
Fair enough - you've convinced me! Being able to runtime adjust is of course extremely helping during _development_ of a user-friendlier adaptive system anyhow.

I like the sound of `perftune.h` - it could also go into [`debug.h`](https://github.com/QuEST-Kit/QuEST/blob/devel/quest/include/debug.h) in the interim to there being more performance-tuning specific functionality.

quest/src/api/environment.cpp
```diff
@@ -509,5 +509,16 @@ void getEnvironmentString(char str[200]) {
 }
 
 
+int getQuESTGpuThreadsPerBlock() {
+    QuESTEnv env = getQuESTEnv();
+    return env.isGpuAccelerated? gpu_getNumThreadsPerBlock() : 0;
+}
+
+void setQuESTGpuThreadsPerBlock(const int NEW_TPB) {
+    // just rely on the internal function to throw an error if there's no GPU support compiled
```
TysonRayJones: TODO: validate this is a factor of 32 (and is positive, etc etc)

otbrown:
Should validate TPB is multiple of 32!

quest/src/api/environment.cpp
TysonRayJones: Doc to user: HIP warpsize is 64!

quest/src/api/environment.cpp
```diff
@@ -509,5 +509,16 @@ void getEnvironmentString(char str[200]) {
 }
 
 
+int getQuESTGpuThreadsPerBlock() {
+    QuESTEnv env = getQuESTEnv();
```
TysonRayJones: Note `getQuESTEnv()` is an API function with its own validation, and shouldn't be called internally like this since if its validation throws, it will claim the user called `getQuESTEnv()`, when they actually called `getQuESTGpuThreadsPerBlock()`. 

Should therefore first call
```
    validate_envIsInit(__func__);
```
and subsequently use `globalEnvPtr` directly. 

(I see `getEnvironmentString()` calls `getQuESTEnv()` for some reason, when it too should just use `globalEnvPtr`)

quest/include/environment.h
```diff
@@ -83,6 +83,14 @@ int isQuESTEnvInit();
 QuESTEnv getQuESTEnv();
 
 
+/** @notyetdoced
+ * GPU thread per block control
+ * This is somehow probably the best pre-existing place for this. It only really applies to GPU, because for
+ * OpenMP the user can just export OMP_NUM_THREADS or call omp_set_num_threads.
+ */
+int getQuESTGpuThreadsPerBlock();
```
TysonRayJones: Should we include `Num` somewhere, e.g.
- `getNumQuESTGpuThreadsPerBlock`
- `getQuESTNumGpuThreadsPerBlock`
 
I've so far tried to avoid abbreviating where feasible.

quest/include/environment.h
```diff
@@ -83,6 +83,14 @@ int isQuESTEnvInit();
...
+void setQuESTGpuThreadsPerBlock(const int NEW_TPB);
```
TysonRayJones: (Same `Num`) consideration as for `getQuESTGpuThreadsPerBlock`)

quest/src/api/environment.cpp
TysonRayJones: Should also call
```
validate_envIsInit(__func__);
```

quest/src/api/environment.cpp
```diff
@@ -509,5 +509,16 @@ void getEnvironmentString(char str[200]) {
...
+    return env.isGpuAccelerated? gpu_getNumThreadsPerBlock() : 0;
```
TysonRayJones: Hmm I think this is a pitfall. If `setQuESTGpuThreadsPerBlock()` is permitted in non-GPU mode (and I really believe _it should_ for healthy platform agnosticism), then the user would always get back `0` in lieu of what they had just passed to set. Maybe we should just always return `gpu_getNumThreadsPerBlock()`.

The situation is slightly different to the GPU cache (fetchable by `getGpuCacheSize()` and clearable via `clearGpuCache()`), because that offers no _setter_. Users can always safely call both in non-GPU accelerated mode, and the former will return `0` (which is always correct).

quest/src/api/environment.cpp
```diff
@@ -509,5 +509,16 @@ void getEnvironmentString(char str[200]) {
...
+}
+
+void setQuESTGpuThreadsPerBlock(const int NEW_TPB) {
```
TysonRayJones: `NEW_TPB` -> `numThreadsPerBlock` or `numTPB`, etc

quest/src/api/environment.cpp
TysonRayJones: Maybe a better alternative: add `gpu_isHipCompiled()` in `gpu_config.cpp`, right under `gpu_isCuQuantumCompiled()`, as:
```cpp
bool gpu_isHipCompiled() {
    return (bool) (COMPILE_CUDA && defined(__HIP__));
}
```
Then we can validate explicitly that when GPU-accelerated _and_ we're on HIP, arg must be a multiple of 64, else of 32. This means 32 is required even when _not_ GPU-accelerated; so we make that error message:
> The number of threads per block must be a multiple of 32 (or on AMD GPUs, 64)

quest/src/gpu/gpu_config.cpp
```diff
@@ -41,6 +41,7 @@
     #include "quest/src/gpu/cuda_to_hip.hpp"
 #endif
 
+int numThreadsPerBlock = 128;
```
TysonRayJones: Should give this a `global_` prefix, like [here](https://github.com/QuEST-Kit/QuEST/blob/f1606dafd389e7dfef34bf8b54429d4e43bf8992/quest/src/core/validation.cpp#L1166) (like I failed to do for `hasGpuBeenBound`, oops!)

Further, given it's not accessed anywhere outside `gpu_(g|s)etNumThreadsPerBlock()`, I would move this definition to the `ENVIRONMENT MANAGEMENT` section, just before `gpu_getNumThreadsPerBlock()`.

quest/src/gpu/gpu_config.cpp
```diff
@@ -330,6 +331,24 @@ qindex gpu_getMaxNumConcurrentThreads() {
...
+}
+
+void gpu_setNumThreadsPerBlock(const int newThreadsPerBlock) {
+#if COMPILE_CUDA
+    numThreadsPerBlock = newThreadsPerBlock;
+#else
+    error_gpuQueriedButGpuNotCompiled();
+#endif
+    return;
+}
```
TysonRayJones, line 370: If we permit users to call the corresponding API functions when GPU acceleration is not enabled, then these guards can be removed entirely. I think that's fair/natural, because we certainly shouldn't introduce an API difference between compiling but _not running_ with GPU acceleration.

I would also comment this exception. So this could become:
```cpp
int gpu_getNumThreadsPerBlock() {
    // permitted even when GPU backend not compiled
    return globlal_numThreadsPerBlock;
}

void gpu_setNumThreadsPerBlock(const int newThreadsPerBlock) {
    // permitted even when GPU backend not compiled
    global_numThreadsPerBlock = newThreadsPerBlock;
}
```

quest/src/gpu/gpu_kernels.cuh
```diff
@@ -42,23 +42,19 @@
...
-const int NUM_THREADS_PER_BLOCK = 128;
-
-
 __forceinline__ __device__ qindex getThreadInd() {
     return blockIdx.x*blockDim.x + threadIdx.x;
 }
 
 
-__host__ qindex getNumBlocks(qindex numThreads) {
+__host__ qindex getNumBlocks(qindex numThreads, const int numThreadsPerBlock) {
```
TysonRayJones: I would remove the `const` qualifier since it's asymmetric and a bit counterproductive, because it makes the reader why _why isn't_ `numThreads` const

quest/src/gpu/gpu_subroutines.cpp
```diff
@@ -141,13 +140,14 @@ qindex gpu_statevec_packAmpsIntoBuffer(Qureg qureg, SmallList qubits, SmallList
 #if COMPILE_CUDA || COMPILE_CUQUANTUM
 
     qindex numThreads = qureg.numAmpsPerNode / powerOf2(qubits.size());
-    qindex numBlocks = getNumBlocks(numThreads);
+    const int numThreadsPerBlock = gpu_getNumThreadsPerBlock();
```
TysonRayJones: I think the `const` before `numThreadsPerBlock` in `gpu_subroutines.cpp` are superfluous and only cause reader confusion (they might incorrectly infer that only the 2nd arg of `<<<` needs to be `const`, or something). Context already make the constness obvious

otbrown:
Rather than attempt to post a thumbs up on each comment while on train WiFi, I'll just comment thanks @TysonRayJones here! I'm hoping to give this branch some proper attention on Thursday/Friday.

TysonRayJones:
No prob! They're almost all nits I can sweep through rapidly. You can flag any you disagree with or for which there's nuance, and I can address the remainder next time I'm on QuEST. (No need to co-author the squash with me for my minor and mostly stylistic changes!)

quest/src/gpu/gpu_subroutines.cpp
otbrown: Resolved in 680fdda.

otbrown:
Okay, so I've managed a handful of minor updates on this today 🤦‍♂️ I've kicked back the grand plan to create a perftune.h and will just focus on getting this version polished enough to merge for v4.3

quest/src/api/environment.cpp
otbrown: Resolved in 99c3b3c.

quest/src/api/environment.cpp
otbrown: Resolved in 99c3b3c.

quest/include/environment.h
otbrown: Resolved in 99c3b3c.

quest/include/environment.h
otbrown: Resolved in 99c3b3c.

quest/src/api/environment.cpp
otbrown: Ha, I suspect given the placement of `getEnvironmentString` that I copied the pattern it used 😅 I've updated that as well as the new functions in 99c3b3c.

quest/src/gpu/gpu_config.cpp
otbrown: Resolved in 40a16c1.

quest/src/api/environment.cpp
otbrown: Resolved in fcbd4f4. I've also added `isHipCompiled` to the `printCompilationInfo` reporting in 8c2a65b, because why not. Note that it is possible (albeit probably not recommended) to compile for NVIDIA using HIP, it's right that it remains somewhat distinct from `isGpuCompiled`.

quest/src/gpu/gpu_config.cpp
otbrown: Resolved in 718db7c.

quest/src/api/environment.cpp
otbrown: Fair point! Resolved in 718db7c.

quest/src/api/environment.cpp
otbrown: I won't resolve this thread just yet as I _do_ still need to add documentation!

otbrown:
Getting there -- still need to add tests, documentation, and review some `const`!

quest/src/gpu/gpu_kernels.cuh
otbrown: I already know we have quite different views on `const`-ness 😛 there's no performance benefit here, it's just a defensive style, so I don't see any need to fight for it! Resolved in [97dbf08](https://github.com/QuEST-Kit/QuEST/pull/736/commits/97dbf085f1812db1a01f8354a4217bc5af0918ee).

quest/src/api/environment.cpp
otbrown: CMake var is now documented in 2edb976. Definitely a benefit to a wider performance documentation effort, but I think that should go with the proposed perfTune header, so deferred until then.

TysonRayJones:
I'll fix the validation issues, and move the API function; I think it doesn't belong in `environment.h` which is concerned with managing the `QuESTEnv` type. The most appropriate "long term" location is `debug.cpp`, which contains other controls of QuEST behaviour, such as validation precision, report lengths, etc. However, if we believe this facility/interface will be changed/extended in the future, we could presently move it into `experimental.h`. Recall that this is "nuisance free"; if we later decide to retain this API without change, moving it from `experimental.h` to `debug.h` has no change to UX. Is that all okay?

docs/cmake.md
```diff
@@ -48,7 +48,7 @@ make
 | `QUEST_DISABLE_DEPRECATION_WARNINGS` | (`OFF`), `ON` | Whether to disable the compile-time deprecation warnings when using the deprecated (v3) API. |
 | `USER_SOURCE_NAMES` | (Undefined), String | The source file for a user program which will be compiled alongside QuEST. `USER_OUTPUT_EXE_NAME` *must* also be defined. |
 | `USER_OUTPUT_EXE_NAME` | (Undefined), String | The name of the executable which will be created from the provided `USER_SOURCE_NAMES`. `USER_SOURCE_NAMES` *must* also be defined. |
-
+| `QUEST_GPU_NUM_THREADS_PER_BLOCK` | (128), Number | The default number of threads per block QuEST will use when offloading to a GPU. *Must* be a multiple of 32. For AMD GPUs this *should* be a multiple of 64. |
```
TysonRayJones: A note on naming: `QUEST_GPU_NUM_THREADS_PER_BLOCK` documented here is inconsistent with the actual parameter name, `QUEST_DEFAULT_NUM_THREADS_PER_BLOCK`. But I think _both_ `GPU` and `DEFAULT` are important to include, so advocate for the third name:
```cpp
QUEST_DEFAULT_GPU_NUM_THREADS_PER_BLOCK
```
This clarifies the threads discussed here are GPU specific (not to be confused with CPU threads), _and_ that this only overrides the default, in the same fashion as `QUEST_DEFAULT_VALIDATION_EPSILON` (which is an env-var).

TysonRayJones: Further, what is the motivation of making this variable a CMake/compile-time constant? The fact it can be runtime overriden means we don't gain any performance benefit. So we can penalty-free _improve_ the flexibility by making this instead an environment variable, just like `QUEST_DEFAULT_VALIDATION_EPSILON`, so that changing it doesn't require recompilation

quest/include/environment.h
```diff
@@ -92,6 +92,14 @@ int isQuESTEnvInit();
 QuESTEnv getQuESTEnv();
 
 
+/** @notyetdoced
```
TysonRayJones: Have to add a newline after command `@notyetdoced`, or the text below gets included in the command's created warning box

quest/include/environment.h
```diff
@@ -92,6 +92,14 @@ int isQuESTEnvInit();
...
+ * GPU thread per block control
+ * This is somehow probably the best pre-existing place for this. It only really applies to GPU, because for
+ * OpenMP the user can just export OMP_NUM_THREADS or call omp_set_num_threads.
+ */
+int getQuESTNumGpuThreadsPerBlock();
+void setQuESTNumGpuThreadsPerBlock(const int newThreadsPerBlock);
```
TysonRayJones: Needs separate TODO doc. We should reference these functions from `initCustomQuESTEnv` (or somewhere similar) to indicate _further control_ of runtime environment is possible

quest/src/api/environment.cpp
```diff
@@ -192,11 +192,12 @@ void printCompilationInfo() {
...
         "compilation", {
-        {"isMpiCompiled",                comm_isMpiCompiled()},
+        {"isOmpCompiled",       cpu_isOpenmpCompiled()},
+        {"isMpiCompiled",       comm_isMpiCompiled()},
         {"isMpiSubCommunicatorCompiled", comm_isMpiSubCommunicatorCompiled()},
-        {"isGpuCompiled",                gpu_isGpuCompiled()},
-        {"isOmpCompiled",                cpu_isOpenmpCompiled()},
-        {"isCuQuantumCompiled",          gpu_isCuQuantumCompiled()},
+        {"isGpuCompiled",       gpu_isGpuCompiled()},
+        {"isHipCompiled",       gpu_isHipCompiled()},
```
TysonRayJones: Great idea to introduce this, but I don't like that the binary is HIP-focused. We can replace `isHipCompiled` with "gpuPlatform", with options "CUDA" or "HIP"

TysonRayJones: I think the binary of `isHipCompiled` would be useful to attach directly to `QuESTEnv` too, like James did for `isMpiGpuAware`

quest/src/api/environment.cpp
TysonRayJones: (Unresolved to remind myself since validation is missing - internal errors used instead)

quest/src/gpu/gpu_config.cpp
```diff
@@ -330,6 +335,29 @@ qindex gpu_getMaxNumConcurrentThreads() {
...
+void gpu_setNumThreadsPerBlock(const int newNumThreadsPerBlock) {
+    if (gpu_isHipCompiled()) {
+        // number of threads per block should be a multiple of 64
+        if (newNumThreadsPerBlock % 64)
+            error_gpuBadNumThreadsPerBlock();
+    } else {
+        // number of threads per block should be a multiple of 32
+        if (newNumThreadsPerBlock % 32)
+            error_gpuBadNumThreadsPerBlock();
+    }
```
TysonRayJones: Turn this into validation in `vlalidation.cpp`, but it's nice to keep this as an internal error check; compact into e.g.
```cpp
void assert_gpuNumThreadsPerBlockIsWarpDivisible(int numThreadsPerBlock) {
    int warpSize = gpu_isHipCompiled()? 64 : 32;
    if (numThreadsPerBlock % warpSize != 0)
        raiseInternal("whatever msg is currently in error_gpuBadNumThreadsPerBlock()");
}
```

tests/unit/environment.cpp
```diff
@@ -185,6 +185,40 @@ TEST_CASE( "getQuESTEnv", TEST_CATEGORY ) {
...
+        // set it back to the original and confirm that also worked
+        REQUIRE_NOTHROW(setQuESTNumGpuThreadsPerBlock(QUEST_DEFAULT_NUM_THREADS_PER_BLOCK));
+        REQUIRE(getQuESTNumGpuThreadsPerBlock() == QUEST_DEFAULT_NUM_THREADS_PER_BLOCK);
+
+    }
+
+    SECTION( LABEL_VALIDATION ) {
+
+        // The way the error-handling currently works, Catch2 can't catch these (ironically)
+        // but leaving them in case we ever update the way errors are done.
```
TysonRayJones: It'll catch when internal errors changed to `validation.cpp` validators

CMakeLists.txt
```diff
@@ -183,6 +183,20 @@ option(
...
+  CACHE
+  STRING
+  "The default number of threads per block QuEST will use when offloading to a GPU. Set to 128 by default. Must be a multiple of 32."
+)
+mark_as_advanced(QUEST_GPU_NUM_THREADS_PER_BLOCK)
+
+math(EXPR quest_tpb_remainder "${QUEST_GPU_NUM_THREADS_PER_BLOCK} % 32")
+if ((NOT (quest_tpb_remainder EQUAL 0)) OR (QUEST_GPU_NUM_THREADS_PER_BLOCK LESS 32))
+    message(FATAL_ERROR "QUEST_GPU_NUM_THREADS_PER_BLOCK must be a multiple of 32. QUEST_GPU_NUM_THREADS_PER_BLOCK=${QUEST_GPU_NUM_THREADS_PER_BLOCK}.")
+endif()
```
TysonRayJones: Remove if we switch to env-var

CMakeLists.txt
```diff
@@ -500,6 +514,7 @@ set(QUEST_COMPILE_MPI ${QUEST_ENABLE_MPI})
 set(QUEST_COMPILE_SUBCOMM ${QUEST_ENABLE_SUBCOMM})
 set(QUEST_COMPILE_CUQUANTUM ${QUEST_ENABLE_CUQUANTUM})
 set(QUEST_INCLUDE_DEPRECATED_FUNCTIONS ${QUEST_ENABLE_DEPRECATED_API})
+set(QUEST_DEFAULT_NUM_THREADS_PER_BLOCK ${QUEST_GPU_NUM_THREADS_PER_BLOCK})
```
TysonRayJones: Remove if we switch to env-var

quest/src/api/environment.cpp
```diff
@@ -509,23 +511,36 @@ void reportQuESTEnv() {
...
     snprintf(str, 200, "CUDA=%d OpenMP=%d MPI=%d userOwnsMPI=%d threads=%d ranks=%d cuQuantum=%d gpuDirect=%d",
-        env.isGpuAccelerated,
-        env.isMultithreaded,
-        env.isDistributed,
-        env.userOwnsMpi,
+        globalEnvPtr->isGpuAccelerated,
+        globalEnvPtr->isMultithreaded,
+        globalEnvPtr->isDistributed,
+        globalEnvPtr->userOwnsMpi,
         numThreads,
```
TysonRayJones, line 539: can print numGpuThreadsPerBlock here (and rename threads to cpuThreads)

tests/unit/environment.cpp
```diff
@@ -185,6 +185,40 @@ TEST_CASE( "getQuESTEnv", TEST_CATEGORY ) {
 }
 
 
+TEST_CASE( "QuESTNumGpuThreadsPerBlock", TEST_CATEGORY ) {
```
TysonRayJones: must be the API function name

tests/unit/environment.cpp
```diff
@@ -185,6 +185,40 @@ TEST_CASE( "getQuESTEnv", TEST_CATEGORY ) {
...
+
+    SECTION( LABEL_CORRECTNESS ) {
+        // Check that it initially matches the compile time value
+        // stored in config.h
+        REQUIRE(getQuESTNumGpuThreadsPerBlock() == QUEST_DEFAULT_NUM_THREADS_PER_BLOCK);
```
TysonRayJones: nit: spaces around macro invocation, i.e. `REQUIRE( x )`

TysonRayJones (commented):
(I'll address these in #767)

quest/src/api/environment.cpp
TysonRayJones: (holding off on this; everything else reported here is [at least intendedly] fixed over runtime. The semi-exception is the reported num-threads, but users cannot modify that via QuEST: instead they'd need to use the OMP runtime. we can return and add gpuNumThreads when we sort out making cpuNumThreads runtime variable)

docs/cmake.md
otbrown: See reply on #767 -- the flexibility is a convenience, the compilation is the important bit 😅

tests/unit/environment.cpp
otbrown: Fair, just need to split it into "set" and "get" in that case.

tests/unit/environment.cpp
TysonRayJones: ^ that was such a regrettable choice of variable

quest/src/api/environment.cpp
TysonRayJones: Didn't add this in the end - reserved for #769

PR #737, opened 2026-04-27
Fix applyMultiStateControlledSqrtSwap argument list

D-Exposito:
Minimal change to match argument list of `applyMultiStateControlledSqrtSwap` on implementation and header files. 

On the header file, `applyMultiStateControlledSqrtSwap`  had an extra argument `numControls` which is not needed when taking `std::vector<int>` arguments. This difference caused compilation errors.

```cpp
void applyMultiStateControlledSqrtSwap(Qureg qureg, vector<int> controls, vector<int> states, int qubit1, int qubit2) {...}
// VS
void applyMultiStateControlledSqrtSwap(Qureg qureg, vector<int> controls, vector<int> states, int numControls, int qubit1, int qubit2) { ... }
```

Let me know if a pull request is not the correct protocol for fixing something as small as this.


otbrown:
Ah yes, I see the declaration doesn't match the definition at https://github.com/QuEST-Kit/QuEST/blob/9d7618d7263e3bfba433b88cf1eac0647f08fa0a/quest/src/api/operations.cpp#L768

Good catch, thanks! Assuming tests pass, I'll merge into devel as we're only a month out from a new minor release.

otbrown:
Hi @D-Exposito -- in the end I've migrated your commit to a branch based on devel to make my life easier. It looks like I've managed to preserve the authorship record okay, so I'll close this one.

Your commit will now be merged in #738!

TysonRayJones:
Wew nice catch! @D-Exposito do you wish to be included in the QuEST patch authors list? If so, please share a name (or we can just use your account handle) to be listed e.g. [here](https://github.com/QuEST-Kit/QuEST#acknowledgements). This is a microscopic diff of course, but pixels are free!

D-Exposito:
Hi @TysonRayJones ! I feel it's quite undeserved, but it would be cool to appear if you offer. My full name is Daniel Expósito Patiño. I'll keep working with QuEST, maybe I will find other ways to contribute :)

TysonRayJones:
Added in ac2c728fef7468e999d426c5028d3b325141be56 🎉

PR #742, opened 2026-05-05
CMake warn when non-release build

TysonRayJones:
Warn users about potential performance loss when not explicitly specifying `Release`, and especially when using a multi-config generator which requires build-type specification (as per the [doc](https://github.com/QuEST-Kit/QuEST/blob/main/docs/compile.md#optimising)) - since this just bit me pretty bad on Windows! 😠

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -187,6 +187,7 @@ message(STATUS "Disabling of deprecated API warnings is turned ${DISABLE_DEPRECA
 option(INSTALL_BINARIES "Whether to include example and user binaries in the install." OFF)
 
 
+
 # ============================
 # Validate options
 # ============================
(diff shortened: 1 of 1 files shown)


TysonRayJones:
@otbrown Any objections to this idea? (I worry about adding warnings to the build process generally).

This will not affect single-config (i.e. MacOS / Linux) users who leave `CMAKE_BUILD_TYPE` unspecified, since it defaults to `Release` as per your 35de0e4e4b9add9cbfbfa682f73651ccc5ad19c0. Single-config users will only see the warning when explicitly overriding `CMAKE_BUILD_TYPE` to a non-release mode (overriding to `""` will default to `Release`).

In contrast, multi-config users (e.g. wretched Windows users) will _always_ see a message at config time, reminding them to subsequently build with `--config Release`. They will also see the message when rebuilding after making a config change too, since CMake auto reconfigs.

It's my opinion this is all good/fine, but making sure as a CMake noobie!

otbrown:
The spirit is fine! I'm just looking at [config generators](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:CONFIG) which should let us do it more neatly and capture both single and multiconfig users (although isn't actually working when tested currently...). 

I'll push an alternative version using that if I can get it working, if not, squash and merge!

otbrown:
Okay, after some footering around I have concluded that Generator expressions will _not_ help here. I will update to include `RelWithDebInfo`, and possibly case insensitivity.

otbrown:
@TysonRayJones ignoring my naive and failed attempt to suppress the stupid EOF diff, which I suspect is a LF/CRLF issue:

I have added something to deal with case insensitivity of CMAKE_BUILD_TYPE and included `RelWithDebInfo` in the "do not warn" category, as that should generally be equivalent to `-O3 -g`. Assuming I haven't somehow broken anything else, I'm happy for you to merge!

In theory we ought to do something similar where we check the build type before adding `-Ofast`, but seeing as that's all about to be removed anyway, I propose we leave it.

Edit: I had broken it 🤦‍♂️ _should_ be fixed now

TysonRayJones:
Good catches! Confirmed everything looks good in manual testing on Windows and Mac 🎉

PR #743, opened 2026-05-10
Replace vector<int> with stack-list

TysonRayJones:
This is to circumvent the `std::vector` performance overheads visible in few-qubit simulation (responsible for a performance regression from v3; see #720), and also so that qubit lists can be passed directly to CUDA kernels without conversion (as explored in #739).

### DIFF
--- a/quest/src/gpu/gpu_thrust.cuh
+++ b/quest/src/gpu/gpu_thrust.cuh
@@ -37,6 +37,7 @@
 #include "quest/src/core/errors.hpp"
 #include "quest/src/core/bitwise.hpp"
 #include "quest/src/core/constants.hpp"
+#include "quest/src/core/small_list.hpp"
 #include "quest/src/core/utilities.hpp"
 #include "quest/src/core/randomiser.hpp"
 #include "quest/src/core/fastmath.hpp"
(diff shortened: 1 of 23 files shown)


TysonRayJones:
@otbrown Do you have any thoughts about [`small_list.hpp`](https://github.com/QuEST-Kit/QuEST/blob/40b2a7418d80c53ef0f036a9384b5d4b96f4ae33/quest/src/core/small_list.hpp)? It almost seamlessly replaces `std::vector<int>` in the QuEST middle-end, and speeds up few-qubit simulation (eliminating all vector/heap allocs) without giving up copy-by-value safety. 

Just like the `(c|g)pu_qcomp` overhaul, it gives us two birds with one stone, benefitting both CPU and GPU simulation; James' optimisations in #739 can (hopefully!!) leverage the list, so kernels can receive `SmallList` directly with no conversion layer.

otbrown:
Philosophically I hate it, but it's correct and it's faster, so have at it! I work in HPC centre, not a philosophy... place

TysonRayJones:
Hehe I share the feeling that it's regrettable, but at least meshes nicely with James' GPU optimisations. I note the move to `std::vector<int>` to `SmallList` above **_does improve_** few-qubit performance, but is _**not sufficient**_ to reach v3 performance. I will add the final optimisations (mostly avoiding re-initialisation of internal objects) in another PR for cleanliness.

quest/src/gpu/gpu_thrust.cuh
```diff
@@ -64,12 +65,22 @@
...
  * to the data (d_vec.data()) can be cast into a raw pointer
  * and passed directly to CUDA kernels (though qcomp must be
- * reinterpreted to gpu_qcomp)
+ * reinterpreted to gpu_qcomp).
  */
 
 
 using devints = thrust::device_vector<int>;
 
+devints getDevInts(SmallList h_list) {
```
TysonRayJones, line 74: This should be made redundant by James' work

TysonRayJones:
Tests pass in all configurations, _except_ that the cuQuantum backend has not been run, which (for convenience) I aim to do via the paid CI runner before the v4.3 release is finalised.

Issue #751, opened 2026-05-19
Insert `QuEST` into debug API


TysonRayJones:
The [debug](https://github.com/QuEST-Kit/QuEST/blob/main/quest/include/debug.h) API functions...
- `setSeeds()`
- `setSeedsToDefault()`
- `getSeeds()`
- `getNumSeeds()`
- `setInputErrorHandler()`
- `setValidationOn()`
- `setValidationOff()`
- `setValidationEpsilonToDefault()`
- `setValidationEpsilon()`
- `getValidationEpsilon()`
- `setMaxNumReportedItems()`
- `setMaxNumReportedSigFigs()`
- `setNumReportedNewlines()`
- `setReportedPauliChars()`
- `setReportedPauliStrStyle()`
- `getGpuCacheSize()`
- `clearGpuCache()`
- `getEnvironmentString()`

are named irksomely generically, and are likely to collide with user symbols (`setSeeds()`?! What was I thinking?!), or those of other libraries/dependencies.

In contrast, the [environment](https://github.com/QuEST-Kit/QuEST/blob/main/quest/include/environment.h) API functions...
- `initQuESTEnv()`
- `initCustomQuESTEnv()`
- `finalizeQuESTEnv()`
- `syncQuESTEnv()`
- `reportQuESTEnv()`
- `isQuESTEnvInit()`
- `getQuESTEnv()`
conveniently contain `QuEST`, disambiguating from user namespace.

We should rename the debug functions to also include `QuEST`. The remaining API is fine, since containing either the name of a QuEST type (e.g. `Qureg`, `CompMatr` `PauliStr`, ...) or are an unambiguously "QuESTy" operation (`applyHadamard`, `calcFidelity`, ...)

While doing this, we may also consider renaming potentially ambiguous preprocessors (such as `FLOAT_PRECISION`), in addition to the CMake flags already nominated for rename in #709 .

TysonRayJones:
@otbrown Were any major API breaks planned for v4.3? (I consider the Trotter `permuteTerms` argument to be a very minor break). Would bumping only a minor version (to v4.3, rather than v5) be inappropriate with the notable API break proposed in this issue?

otbrown:
Hallo, it's an interesting question!

[...]

Issue #709, opened 2026-01-30
Prefix QuEST CMake options


vaferreiQMT:
Currently QuEST's CMake variables, such as `BUILD_EXAMPLES` and `ENABLE_TESTING` are defined globally with no prefix. This is fine when used in isolation but can become a bit of a hassle when QuEST is used as a dependency/ within a larger project.
These options should be prefixed to avoid clashes (e.g. `QUEST_BUILD_EXAMPLES`).

otbrown:
Hi Vasco -- no argument from me! I've plans to revisit the MPI in QuEST over the next couple of weeks, so I can create a PR for this at the same time if you haven't already.

TysonRayJones:
Have prepared a debug API rename in #752, and environment var rename in #755.

Below I compile the CMake vars and preprocessors. I want to clean them up beyond just inserting `QUEST` into them, while we're breaking API. I have indicated further renaming with `# !` - I'll come back and add the nominated rename with an explanation (about to disappear to shid my pants watching Obsession)

Here are the below user-facing CMake variables:
```python
# relevant only to single-exec mode
USER_SOURCE    # !
OUTPUT_EXE     # !

# relevant only to installation
LIB_NAME            # !
VERBOSE_LIB_NAME    # !
INSTALL_BINARIES

# accuracy hyper-params
FLOAT_PRECISION

# backend compilation
ENABLE_MULTITHREADING    # !
ENABLE_DISTRIBUTION      # !
ENABLE_CUQUANTUM
ENABLE_CUDA
ENABLE_HIP

# additional compilation
BUILD_EXAMPLES
ENABLE_TESTING         # !
DOWNLOAD_CATCH2

# extra API
ENABLE_DEPRECATED_API
DISABLE_DEPRECATION_WARNINGS
```

Then we have internal CMake vars / funcs - can they collide with user-given CMake vars?
```python
# potentially ambiguous
default_build_type
NUMA_AWARE                # !

# disambiguated to QuEST
setup_quest_rpath
QuEST_INSTALL_CONFIGDIR     # !

```

We also have the below internal preprocessors; could they foreseeably collide with user-set preprocessors, triggering the compile-time error within `config.h.in`?
```python
# one-to-one with CMake var
FLOAT_PRECISION
INCLUDE_DEPRECATED_FUNCTIONS
DISABLE_DEPRECATION_WARNINGS
NUMA_AWARE

# informed by CMake var
COMPILE_OPENMP
[...]

PR #752, opened 2026-05-19
Insert QuEST into debug API names

TysonRayJones:
as per #751. Note there's no need to actually review the horrifying diff of course :^) just the idea! Fine to defer this PR for a future release it it's too great an API break for v4.3 (and we wished to defer v5).

Note this PR does _not_ also rename the preprocessors and CMake vars, nominated by #709, which should be isolated to another PR.

TODO:
- [x] update doxygen URLs


TysonRayJones:
**BEWARE** that before merging, we must update the URLs in doc hyperlinks, which contain hashes of the changed function signatures. The new URLs can be learned via local Doxygen generation. Search for sub-URL: `https://quest-kit.github.io/QuEST/group__debug`

TysonRayJones:
Let's also _**not merge this**_ until a PR with similar changes to the CMake vars are ready (so that, just in case we defer this until next release, we will not be violently breaking API _twice_)

PR #762, opened 2026-05-29
Cleanup custom MPI flow

TysonRayJones:


quest/src/core/validation.cpp
```diff
@@ -1482,6 +1497,60 @@ void validate_gpuIsCuQuantumCompatible(const char* caller) {
...
+
+    // (A) If the user does not declare ownership of MPI, they are forbidden to initialise it
+    if (!userOwnsMpi)
+        assertThat(!isMpiInit, report::QUEST_OWNED_MPI_WAS_PRE_INIT, caller);
+
+    // (B) If QuEST is instructed not to use distribution, we must demand the user is not
+    // using MPI, because we internally consult comm_isInit() to detect QuEST distribution
+    // in many functions, and that will give a false positive when the user inits MPI directly. 
+    if (!useDistrib)
+        assertThat(!isMpiInit, report::QUEST_IS_NON_DISTRIBUTED_BUT_MPI_WAS_INIT, caller);
```
TysonRayJones: @otbrown is this limitation okay? It is: when QuEST is **_compiled_** with MPI, but distribution is _**disabled**_ at runtime (during `initCustom(Mpi...)QuESTEnv`), we _**forbid**_ the user to use MPI themselves. This is due to a relatively minor technical limitation; QuEST internally consults `comm_isInit()` (which will detect the user's `MPI_Init()` call) in many places as a proxy for "is QuEST distributed?".

A current workaround for such users is to re-compile QuEST _without_ MPI, in which case `comm_isInit()` decouples from MPI. And alas currently, a user initailising MPI _after_ a non-distributed (but MPI-compiled) QuEST initialisation will corrupt QuEST - it will survive `initQuESTEnv` validation but later will be tricked by `comm_isInit()`.

**EDIT:** during writing this, I've convinced myself that it should be valid/legal for users to compile QuEST with MPI, but runtime disable QuEST distribution, and thereafter discretionarily call `MPI_Init`. Corrupting QuEST is a major pitfall. I'll fix!

TysonRayJones: Limitation removed in #765 as discussed in meeting

TysonRayJones:
This must not be merged until #765 is merged into this

PR #764, opened 2026-05-30
(invalid) define `env_isDistributed()`

TysonRayJones:
This was an attempt to avoid consulting `comm_isInit()` internally, which only indicates MPI is active, and not whether QuEST is allowed to use it! (QuEST itself may be launched non-distributed within a user-owned MPI environment). This was to relax the current restriction highlighted [here](https://github.com/QuEST-Kit/QuEST/pull/762#discussion_r3327028466).

This was already an ugly attempted solution because of the nuisance of not being able to expose internal functioj `env_isDistributed()` within the user-facing `environment.h` header...

It is, in fact a totally inadequate / non-functioning solution! This is due entirely to a **_single_** function; `gpu_areAnyNodesBoundToSameGpu()`. This is the _only_ function in all of QuEST which needs to perform an MPI communication _before_ the validation within `initQuESTEnv` has been completed (indeed, it's invoked by validation), in scenarios where the validation is _not_ in the process of failing. This precludes us from calling `env_isDistributed()` within it, since the `QuESTEnv` is still `nullptr`, and the "in-process" `initQuESTEnv` call has nowhere to indicate that it _intends_ to distribute QuEST. More info in the `gpu_config.cpp` diff.

Consistent with my typical luck, this function is defined in `gpu_config.cpp`, which was shown at the _**very bottom**_ of my IDE's search results for files containing `comm_isInit()`. Ergo, it was the _very last_ function to be updated in this refactor, maximally prolonging my ordeal.

I've made this PR anyway for reference while re-attempting a solution. Thankfully, a cleaner and more robust design exists (consulting of `mpiCom == NULL` from within `comm_config.cpp`)

### DIFF
--- a/quest/src/api/channels.cpp
+++ b/quest/src/api/channels.cpp
@@ -28,6 +28,19 @@ using std::vector;
 
 
 /*
+ * EXTERNAL FUNCTIONS
+ *
+ * which are regrettably extern'd, rather than included
+ * in a header, because they are defined within /api/,
+ * within which the headers are user-visible. Gross!
+ */
+
+
+extern bool env_isDistributed();
+
+
+
+/*
  * PRIVATE UTILITIES
  */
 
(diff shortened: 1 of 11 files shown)


TysonRayJones:
Superseded by #765

PR #765, opened 2026-05-30
Permit usage of MPI when QuEST is non-distributed

TysonRayJones:
- added comm_isActive to indicate whether QuEST is using MPI (which is distinct from whether MPI itself is initialised)
- renamed comm_isInit to comm_isMpiInit, since it queries MPI directly/globally, and when true, does not indicate whether QuEST is actually using MPI
- record isMpiUserOwned within comm_config.cpp, since failed-validation must not kill user-owned MPI, and it must know user-ownership before QuESTEnv succeeds/records it (because validation can fail DURING QuESTEnv initialisation)
- explicitly divided (through doc) comm_config.cpp into things which query MPI globally, and thinks which query only QuEST's MPI env/communicator -


TysonRayJones:
@otbrown This is a substantial change/diff and would be a tedium to review, so don't feel obliged to look at the code! Just checking that the changes as summarised in the text above are conceptually okay / not objectionable 🙏

otbrown:
Yep this all sounds fine to me!

PR #767, opened 2026-05-31
Cleanup of "Set gpu tbp"

TysonRayJones:


TysonRayJones:
Above has no changes; just merging (with conflict resolution) of #765, since it contains the new `experimental.h`

quest/src/gpu/gpu_kernels.cuh
```diff
@@ -301,7 +301,11 @@ __global__ void kernel_statevec_anyCtrlFewTargDenseMatr(
     // must be strictly through compile-time-known indices, otherwise it will auto-
     // spill to local memory). Hence, this _subA() function is not a subroutine 
     // despite some logic being common to non-compile-time _subB(), and hence
-    // why the loops below are explicitly compile-time unrolled
+    // why the loops below are explicitly compile-time unrolled. Beware that when
+    // numThreadsPerBlock is increased from 128, this kernel will still behave
+    // correctly, but privateCache below will spill over into local memory at a
+    // performance penalty for NumTargs <= 5, with spillage occurring for fewer
+    // NumTargs as numThreadsPerBlock increases.
```
TysonRayJones, line 308: @otbrown just flagging this!

quest/include/environment.h
```diff
@@ -35,16 +35,22 @@ extern "C" {
...
+    bool isCuQuantumEnabled;
 
     // deployment configurations which can be changed via environment variables
-    int isGpuSharingEnabled;
+    bool isGpuSharingEnabled;
+
+
+        // TODO: we are attaching this for now, but we should, in a subsequent PR,
+        // attach all the important info to the QuESTEnv, for user consumption!
+        bool isHipCompiled;
```
TysonRayJones: Must deal with this in another PR

TysonRayJones: Noted in #768

TysonRayJones:
@otbrown Note this diff is polluted when a merge from #765, so that I could make use of `experimental.h`. I wouldn't bother looking at the diff anyway, it's a doozy! Instead, I flag the main changes:
- moved `setQuESTNumGpuThreadsPerBlock()` (and the getter) out of `environment.cpp`, and into `experimental.cpp` (in the future, they're appropriate to move to `debug.cpp`)
- replaced the cmake `QUEST_DEFAULT_NUM_THREADS_PER_BLOCK` variable with an environment variable `QUEST_DEFAULT_NUM_GPU_THREADS_PER_BLOCK` (to be consistent with e.g. `QUEST_DEFAULT_VALIDATION_EPSILON`)
- added user-input validation
- documented `setQuESTNumGpuThreadsPerBlock` (and the new env-var)
- replaced the unit tests

Let me know if that's all fine, and I'll do the merging (since I'll need to merge upstream first then solve conflicts)

otbrown:
Thanks @TysonRayJones!

Sounds good to me, _except_ the ability to "bake" this in at compile time is important and should not be lost. This value is architecture dependent, and therefore a facility maintaining a central install of QuEST may wish to set it based on some profiling that they do on their specific hardware. This can then be made available to all users by compiling it in to the library.

The ability to set it at run time was really intended purely as a convenience for the person doing the profiling (*cough* @JPRichings) could iterate through values while doing that profiling without having to recompile every time.

I think the environment variable is a reasonable way to do the runtime setting, but we do still need the compile time setting.

For balance I should note that we _could_ achieve the same with just the environment variable -- we would need to either a) document to users that they should always export the env var before running QuEST, or b) export it for them in the module load. Option a) has obvious problems, but b) would be fine I think. Still, I think the preprocessor variable seems like the better place semantically to set a value _for everyone_. 

Final note is I prefer `perfTune.cpp` to `debug.cpp` for this value and others like it since it's really about tuning for the hardware, not finding or fixing a bug! Not relevant now however.

PR #778, opened 2026-06-04
Add function to sample outcomes from a Qureg

D-Exposito:
Hi!

I added a `sampleQureg` function to obtain multiple random outcomes from the distribution of a `Qureg`. The function is based on `applyMultiQubitMeasurementAndGetProb`, but drawing as many outcomes from the distribution as the user selects (stored as counts on a `std::unordered_map`) and without projecting the `Qureg` afterwards, so information can be extracted at intermediate points during the simulation.

Feel free to reject the pull request if this doesn't mesh well with the philosophy of `QuEST`, as I suspect that the lack of a function of the sort might be intentional.

Best,
Daniel

### DIFF
--- a/quest/include/operations.h
+++ b/quest/include/operations.h
@@ -27,6 +27,7 @@
 
 #ifdef __cplusplus
     #include <vector>
+    #include <unordered_map>
 #endif
 
 
@@ -2342,6 +2343,12 @@ qindex applyMultiQubitMeasurementAndGetProb(Qureg qureg, std::vector<int> qubits
 /// @cppvectoroverload
 /// @see applyForcedMultiQubitMeasurement()
 qreal applyForcedMultiQubitMeasurement(Qureg qureg, std::vector<int> qubits, std::vector<int> outcomes);
+
+/// @notyettested
+/// @notyetvalidated
+/// @notyetdoced
+std::unordered_map<int, int> sampleQureg(Qureg qureg, int* qubits, int numQubits, int shots);
+
 
 
 #endif // __cplusplus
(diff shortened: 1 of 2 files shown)


TysonRayJones:
Hi Daniel,

I have so far spent my career stunned and disappointed by people using a statevector simulator to perform _sampling_; that's an approach for _estimating_ the probability distribution, which QuEST is already directly giving you! Whatever question you were going to answer with the samples has _already been answered_, and to a _greater accuracy_, with the output probabilities! It's an epistemic fallacy akin to asking an all-knowing-superintelligence which page of your encyclopedia you should consult.

As your diff shows, it's also _very easy_ for a user to obtain samples from the probabilities, and so possibly not something worth supporting in QuEST. It would be _more_ (but still not very) worthwhile to support _efficient_ sampling, by parallelising the `shots` loop, to make use of QuEST's hardware accelerated backends in the limit of many shots, which is much harder for users. But humorously, as you increase the number of shots, the sample distribution approaches the distribution of the probabilities - so you're better of just using the probabilities! And again, the question being answered with the shots can still always be answered _absolutely precisely_ from the probabilities directly.

I think users attempting to perform sampling are better off being shown the formula for calculating expectation value and variance. While not appropriate for a QuEST function, it _would_ be a welcome [example program](https://github.com/QuEST-Kit/QuEST/tree/main/examples/extended). I can envision demonstrating sampling, and comparing the sample statistics to a trivial calculation of the expected value and variance. Still humorously circular, but a good chance to teach this to the user!

D-Exposito:
Hi Tyson,

Sorry to disappoint you. The problem surely is that I'm trying to fit QuEST into the shape of a quantum computing simulator, where one wouldn't expect a QPU to return a probability distribution directly. Besides that, I'm now unhappy about not collapsing the state after sampling, as an user could be tempted to "measure for free".

Daniel

Issue #747, opened 2026-05-11
Add `Qureg` checkpointing.


TysonRayJones:
> [!NOTE]
> For all development, please work out of the [`devel`](https://github.com/QuEST-Kit/QuEST/tree/devel) branch

We should add a facility for _checkpointing_; writing the contents of a `Qureg` to a file, so it may be later parsed and loaded into another `Qureg`. This is useful for HPC jobs vulnerable to timeout or other failures - an evolving `Qureg` can be occasionally written to disk and resumed in a subsequent process.

Implementing checkpointing involves defining two new API functions within [`qureg.cpp`](https://github.com/QuEST-Kit/QuEST/blob/devel/quest/src/api/qureg.cpp):
```
void saveQuregToFile(Qureg qureg, const char* fn);
```

```
Qureg createQuregFromFile(const char* fn);
```

The first function (`saveQuregToFile`) should record relevant `Qureg` parameters to the file, like the number of qubits, and whether the `Qureg` is a statevector or density matrix, in addition to the full set of amplitudes. It should _not_ record incidental deployment information (like whether `Qureg` is multithreaded), nor any property which can be inferred from the others (like `numAmps`). 

The second function (`createQuregFromFile`) should parse the file, create a `Qureg` of the specified size, and populate its amplitude buffers with the file contents.

The _challenge_ of checkpointing is performing all this without excessive memory usage, and handling the various `Qureg` deployment schemes, like GPU-acceleration and distribution. Checkpointing should be implemented via integrating [`ADIOS2`](https://github.com/ornladios/ADIOS2). This requires optionally compiling and linking `ADIOS2`, when a CMake variable such as `ENABLE_CHECKPOINTING=1` is specified, exposing the above two functions.

ashmitjsg:
Hey @TysonRayJones, I opened #780 implementing this.

`saveQuregToFile()` and `createQuregFromFile()` are added to `qureg.cpp`, behind a new CMake option `ENABLE_CHECKPOINTING` (requiring ADIOS2). The file records only the `Qureg` dimension and amplitudes - never deployment fields - so a `Qureg` can be restored under a different deployment than it was saved with. Amplitudes are written as an ADIOS2 global array with each node contributing only its local slice, so it streams without excessive memory and is distribution-ready; GPU state is synced to host before writing and back after reading.

[...]

PR #780, opened 2026-06-04
feat: add Qureg checkpointing via ADIOS2 (#747)

ashmitjsg:
## Summary

Implements `Qureg` checkpointing (issue #747): two new API functions for writing a `Qureg` to disk and restoring it later.

```c
void  saveQuregToFile(Qureg qureg, const char* fn);
Qureg createQuregFromFile(const char* fn);
```

This is useful for long-running HPC jobs vulnerable to timeout or failure - an evolving `Qureg` can be periodically written to disk and resumed in a later process.

## Design

Following the approach suggested in the issue, checkpointing is built upon [ADIOS2](https://github.com/ornladios/ADIOS2) and gated behind a new CMake option `ENABLE_CHECKPOINTING` (OFF by default), so the ADIOS2 dependency is only required when the feature is requested.

- **What is saved:** only the `Qureg` dimension (`numQubits`, `isDensityMatrix`) and the full amplitude set. Incidental deployment fields (multithreading, GPU-acceleration, distribution) are *not* saved, nor are derivable fields (`numAmps`, etc). A `Qureg` saved by one deployment can therefore be restored by any other - `createQuregFromFile()` creates the `Qureg` with automatically chosen deployments and a precision marker (`sizeof(qreal)`) guards against restoring into a mismatched-precision build.
- **Amplitudes:** stored as an ADIOS2 global array of interleaved `(real, imag)` reals (`reinterpret_cast` of the contiguous `qcomp` buffer), keeping the format agnostic to precision and to ADIOS2's complex-type support.
- **Memory / deployments:** each node writes/reads only its local slice (`start = 2·rank·numAmpsPerNode`, `count = 2·numAmpsPerNode`) of the global array, so the implementation streams without excessive memory and is distributed-ready. GPU-resident state is synced to host before writing (`syncQuregFromGpu`) and back after reading (`syncQuregToGpu`).
- **Not compiled:** calling either function in a build without `ENABLE_CHECKPOINTING` raises a clear validation error (rather than failing to link), via `validate_quregCheckpointingIsCompiled`.

The new API functions live in the existing C-and-C++-agnostic partition of `qureg.h` (they pass no `qcomp` by value, so remain C-ABI-safe).

## Scope

This first pass targets and is verified for **CPU, single-node** (the ADIOS2 build used here has MPI off). The code is written deployment-agnostically against the global-array abstraction, so enabling distribution (rebuild ADIOS2 + QuEST with MPI) and GPU should work unchanged; I'm happy to extend/verify those if preferred.

## Testing

- New `tests/unit/checkpoint.cpp` (guarded by `ENABLE_CHECKPOINTING`): statevector and density-matrix round-trips assert the restored `Qureg` matches dimension and amplitudes. `./tests/tests "[checkpoint]"` → **all assertions pass** (CPU and CPU+OMP).
- Standalone round-trip check also confirms **bit-exact** restoration (`maxAmpDiff = 0`) for both statevector and density matrix.
- Build verified: `cmake .. -D ENABLE_CHECKPOINTING=ON -D CMAKE_PREFIX_PATH=$HOME/.local` → clean compile + link against `adios2::cxx`.

## Build

```bash
cmake .. -D ENABLE_CHECKPOINTING=ON -D CMAKE_PREFIX_PATH=/path/to/adios2/prefix
cmake --build . --parallel
```
Documented in `docs/compile.md` (new "Checkpointing" section).

## Notes / open questions

- File format is ADIOS2 BP5 (a `.bp` directory). Happy to adjust naming/engine conventions.
- I kept the saved metadata minimal (dimension + precision). If you'd like additional provenance (e.g. a format version field for forward-compatibility), I can add it.

---

### AI-Assisted Contribution Disclosure

I used an AI assistant (Claude) to help explore the QuEST architecture, discuss the design, and review the code and tests. I traced the `Qureg` struct, the `qureg.cpp` / `validation.cpp` patterns, and the amplitude-access and sync routines myself, made the implementation decisions (interleaved-reals storage to dodge ADIOS2's lack of a long-double-complex type; the global-array slice scheme for memory-efficient, distribution-ready I/O; gating via a validation error rather than a link error; where the API and validation belong), and verified all behaviour locally with the tests above plus bit-exact round-trip and build checks. I can explain and stand behind every line.


quest/src/core/validation.cpp
```diff
@@ -1990,6 +1996,35 @@ void validate_quregIsDensityMatrix(Qureg qureg, const char* caller) {
...
+        return;
+
+    // this validation must fire regardless of ENABLE_CHECKPOINTING, so the user
+    // receives a clear error (rather than a linker error) when calling the
+    // checkpointing API in a build which did not compile it
+    #ifdef ENABLE_CHECKPOINTING
+    bool isCompiled = true;
+    #else
+    bool isCompiled = false;
+    #endif
```
TysonRayJones: For this file to see the `ENABLE_CHECKPOINTING` (and other) preprocessors, you must include
```cpp
#include "quest/include/config.h"
```
Presently, the undefined macro will default to `0`, setting `isCompiled=false`, and making this validation always trigger. This makes me suspect you did not compile and run the tests yourself before submitting this PR. Please see test instructions [here](https://github.com/QuEST-Kit/QuEST/blob/devel/docs/launch.md#launch_tests)

TysonRayJones:
For ease of testing, I've made QuEST's build download adios2 when not locally installed

TysonRayJones:
Looks like the CI is successfully downloading adios2, but then _some jobs_ fail to compile! Strangely, the logs (like [this one](https://github.com/QuEST-Kit/QuEST/actions/runs/26990832369/job/79650448682)) don't show an error - compilation just stops. Meanwhile, other jobs on the same platforms (like [this one](https://github.com/QuEST-Kit/QuEST/actions/runs/26990832369/job/79650448685)) compile fine!

Quite irksome, but anyways suggests you should have a go at running your new unit tests yourself, as [guided here](https://github.com/QuEST-Kit/QuEST/blob/main/docs/launch.md#tests). The new CMake provision to download adios2 should make that easier. You can compile the tests and run them locally, even with MPI and few processors, via:
```
mpirun --oversubscribe -np 8 ./tests/tests "[checkpoint]"
```
I'll dig into your implementation once we're assured it compiles and works!

ashmitjsg:
@TysonRayJones You're right, thanks for catching this. I did build and run the tests locally before opening - but via `cmake -DENABLE_CHECKPOINTING=ON`, which set the macro as a raw compile definition on the `QuEST` target. Since `validation.cpp` compiles into that single target, the `-D` reached it in my build, so the `#ifdef` saw it and the `[checkpoint]` round-trip tests passed. The problem is that's the wrong mechanism: it only works for that specific cmake invocation and bypasses QuEST's convention, where compile macros are the single source of truth in `config.h` (and `validation.cpp` doesn't include `config.h`, so by the proper path the macro is undefined and defaults to 0). My mistake - I should have wired it through the config system from the start.

Fixed: added `#cmakedefine01 QUEST_COMPILE_CHECKPOINTING` to `config.h.in`, set it from the CMake option, dropped the direct compile-definition, and switched the sources to `#include "quest/include/config.h"` + `#if QUEST_COMPILE_CHECKPOINTING`. I also linked ADIOS2 to the `QuEST` target (completing the FetchContent block).

Verified locally:
- `cmake -DENABLE_CHECKPOINTING=ON -DQUEST_BUILD_TESTS=ON` -> generated `config.h` has `#define QUEST_COMPILE_CHECKPOINTING 1`; `./tests/tests "[checkpoint]"` passes (6 assertions, 1 test case) across the CPU and CPU+OMP deployments.
- default build (`OFF`) -> `config.h` has `QUEST_COMPILE_CHECKPOINTING 0` and compiles cleanly without ADIOS2.

This was a single-node (non-MPI) build, so I ran the tests serially rather than under `mpirun`. Sorry for the churn.

**Still to do - I'll take these on next:**

- **CI compile stalls.** I believe the "compilation just stops with no error" on some jobs is the runner being OOM-killed during the ADIOS2 FetchContent build (its templates are memory-heavy when compiled in parallel). I'll look into reducing that pressure - e.g. building ADIOS2 with fewer parallel jobs and/or trimming more of its optional components in the FetchContent block - and confirm against the jobs that currently fail.
- **Distributed (MPI) test.** My local build is single-node (no MPI here), so I've only verified the serial CPU / CPU+OMP path. I'll set up MPI locally and run `mpirun --oversubscribe -np 8 ./tests/tests "[checkpoint]"` to exercise the per-rank slice logic (`numAmpsPerNode` / `rank` offsets) end-to-end.
- **GPU path.** The GPU sync (`syncQuregFromGpu` / `syncQuregToGpu`) is in place but I haven't been able to test on a GPU; I'll verify it (or call it out explicitly as untested) so the scope is clear.

I'll report back here with results on each. Thanks for your patience while I get this to a properly-tested state.

ashmitjsg:
@TysonRayJones My last commit closes out the three items I said I'd take on (CI compile stall, distributed MPI test, GPU path). I dug into the failure and ran the tests locally across CPU, MPI, and GPU.

## Root cause of the CI "compilation silently stops"

It was the **CUDA jobs**. The FetchContent block built ADIOS2 with `ADIOS2_USE_CUDA` matched to `QUEST_ENABLE_CUDA`, so on the CUDA runners ADIOS2's own CUDA compile was triggered and stalled. Checkpointing is host-side I/O - amps are synced to host (`syncQuregFromGpu`) before any write, so ADIOS2 never touches device pointers and doesn't need CUDA. Fixed by forcing `ADIOS2_USE_CUDA OFF`.

## Another MPI bug I found while testing distributed

The implementation wasn't actually doing collective I/O:

- `adios2::ADIOS adios;` was constructed without a communicator, and CMake linked the serial `adios2::cxx` target (which doesn't define `ADIOS2_USE_MPI`, so the `ADIOS(MPI_Comm)` constructor isn't even exposed).
- Fixed: link `adios2::cxx_mpi` under MPI and construct `adios2::ADIOS(MPI_COMM_WORLD)`, so each rank collectively writes/reads its slice of one shared file.

## A test-cleanup race

Every rank called `std::filesystem::remove_all` on the shared `.bp`, which nondeterministically threw `cannot remove all: No such file` (and looked like a hang under oversubscription). Fixed: rank-0-only deletion bracketed by `syncQuESTEnv()` barriers. The I/O itself round-trips bit-exact (`maxAmpDiff = 0`).

## Tests run locally - all passing

| Test | Result |
|---|---|
| `[checkpoint]` - serial (non-MPI) build | 6 assertions |
| `[checkpoint]` - `mpirun -np 1/2/4/8` | all pass |
| `[checkpoint]` - your exact `mpirun --oversubscribe -np 8 ./tests/tests "[checkpoint]"` | 6 assertions, "All tests passed" |
| `[checkpoint]` - **GPU** build (`QUEST_ENABLE_CUDA=ON`, sm_75) | exercises `syncQuregFromGpu` / `syncQuregToGpu`, passes |
| GPU numeric sanity (`applyHadamard`) | confirms compute capability set correctly |

So GPU and distributed are now both verified, not just CPU single-node.

## Full unit suite

I also ran the complete suite (not just `[checkpoint]`) across all deployments, using the rigour/runtime env vars documented in `launch.md` - `QUEST_TEST_MAX_NUM_QUBIT_PERMUTATIONS=10` to keep the permutation count tractable locally (the default `0` tests all permutations, which is the CI/cluster-scale config):

- **Single-process** (`./tests/tests`, CPU + CPU+OMP): **All tests passed - 54211 assertions in 277 test cases.**
- **Distributed** (`mpirun -np 2 ./tests/tests`, across CPU / OMP / MPI deployment combinations): running now; I'll post the result here once it finishes.

Nothing I touched is exercised by these beyond linking, but they confirm the `adios2::cxx_mpi` link change and the config plumbing don't disturb the rest of the suite.

ashmitjsg:
@TysonRayJones I re-verified `[checkpoint]` serial / `np 1-8` / `--oversubscribe -np 8` / GPU, and the checkpointing-OFF build compiles clean.

The full *distributed* suite (`mpirun -np 2 ./tests/tests`, all deployments) ran hours on my laptop without finishing, even at reduced rigour, single-process full passed (54211 assertions) and `[checkpoint]` is verified distributed, so I stopped it. Want me to push that full distributed run to completion, or is this enough? Otherwise, ready for review.

TysonRayJones:
Hi Ashmit,
Great work! 🎉 No need to run the full tests - indeed, they're brutal.

 It remains a neat future optimisation to leverage ADIOS2' CUDA awareness, but that's outside the scope of the UH challenge (although if you're interested, we can continue to work with you to ship that). For now, we'll review the current implementation and get back to you!

ashmitjsg:
@TysonRayJones Diagnosed the failing Linux `CKPT` jobs. They were OOM-killed (exit 143, "runner received a shutdown signal") while compiling the bundled ADIOS2, specifically the `EVPath`/`atl`/`ffs`/`dill`/`enet` third-party stack pulled in by ADIOS2's network/staging engines (DataMan/SSC/MHS/SST). It's not a code issue, and it hit every Linux CKPT job, even the plain `OMP` one; Windows runners happened to survive.

Checkpointing only uses the local **BP5 file engine**, so I disabled all the streaming/staging engines (`ADIOS2_USE_DataMan/SSC/MHS/SST/DAOS/MGARD`, plus `Blosc2`). The slimmed ADIOS2 build no longer compiles the EVPath stack. Verified locally: it builds clean and the `[checkpoint]` round-trip passes both serially and under `mpirun -np 2` (25 assertions).

If the Linux runners are still strained after this, the `adios2` matrix axis currently builds ADIOS2 across every CUDA/HIP/cuQuantum combo, even though checkpointing is orthogonal to the GPU backend - I'm happy to restrict `adios2: ON` to the non-GPU combos (base/OMP/MPI) to cut the redundant rebuilds, but that's your call on coverage.

ashmitjsg:
Hi @TysonRayJones, thanks a lot for the CI memory fixes (Free space before ADIOS2 / serial compilation); that nailed the OOM I was hitting on the Linux CKPT jobs. The full matrix is green now. 

Anything else you'd like changed on my end before this can go in?

TysonRayJones:
Hi Ashmit,
It might take us a little longer to get through integration (we will wish to stress-test in lots of different deployment configurations, and potentially seek optimisation on GPU), but you have satisfied the UH requirements. I've assigned the challenge (#747) to you, nice work!! 🎉

ashmitjsg:
Hey @TysonRayJones, thanks for assigning this issue. I want to learn more and get involved in this further in my free time on weekends, so let me know if I can help with the next steps or move towards the near-future optimisation goal of leveraging ADIOS2's CUDA awareness.

PR #782, opened 2026-06-06
lifecycle tutorial

nrdcid:
Added a simple tutorial showing a lightweight entry-point example for new users who want to see the standard QuEST workflow in one small file before exploring more advanced examples.

i.e., `examples/tutorials/min_lifecycle_example.c` shows the basic flow of a small QuEST simulation:

- initialize the QuEST environment
- create a `Qureg`
- initialize the register state
- apply a couple of simple operations
- report the state
- clean up the register and environment

Kept this intentionally minimal and with elaborate comments, did not modify the existing CMake configuration. If preferred, I can follow up by wiring this into the `examples` build system.

### DIFF
--- a/examples/tutorials/min_quest_lifecycle.c
+++ b/examples/tutorials/min_quest_lifecycle.c
@@ -1,0 +1,58 @@
+/** @file
+ * Minimal end-to-end example of a standard QuEST simulation.
+ *
+ * This example demonstrates the typical lifecycle of a QuEST program:
+ *   1. initialise the QuEST environment,
+ *   2. create a quantum register,
+ *   3. initialise its state,
+ *   4. apply operations,
+ *   5. report or calculate results,
+ *   6. destroy allocated objects,
+ *   7. finalise the environment.
+ *
+ * @author Nathan Delcid
+ */
+
+#include "quest.h"
+
+int main(void) {
+
+    // Initialise the QuEST environment. This prepares available hardware
+    // backends such as MPI, GPU acceleration and multithreading if enabled.
+    initQuESTEnv();
+
+    // Create a 2-qubit statevector Qureg.
+    //
+    // Its amplitudes correspond to:
+    //   alpha_0 |00>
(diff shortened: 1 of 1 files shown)


TysonRayJones:
Hi Nathan,

More examples is always more better, thanks very much for this! 🎉  However, doesn't it only meaningfully differ from the existing [`min_example.c`](https://github.com/QuEST-Kit/QuEST/blob/main/examples/tutorials/min_example.c) by only these two lines?
```cpp
    applyHadamard(qureg, 0);
    applyPauliX(qureg, 1);
```

nrdcid:
The main contribution would be the comments. Maybe I can consolidate with `min_example.c`?

TysonRayJones:
That sounds better, but I note that I usually shy from redundant comments, like
```cpp
// Initialise QuEST
initQuESTEnv();

// Create a qureg
Qureg qureg = createQureg(25);

// Apply some operators
applyHadamard(qureg);

// Finalise the QuEST environment.
finalizeQuESTEnv();
```
Such comments are just visual noise when the API function describes precisely what is happening. I much prefer "why" comments, which explain the broader context, as you can see a nice example of [here](https://github.com/QuEST-Kit/QuEST/blob/b983059278ec485f95676bd6df5dad5dec2d5453/quest/src/core/localiser.cpp#L765-L789). Replacing the comments with just the maths doesn't seem worthwhile - it's not a HPC simulator example's job to teach the reader what the Hadamard gate does!

I think `min_example.c` as exists shows the user flow fine. A longer, more meaningful example (e.g. QuEST simulating some canonical quantum algorithm) might be more worthwhile!

nrdcid:
Right right. I mainly did it for myself whilst reading the preprint to translate notation into code. Let me dig a little deeper and see if I can cook something up.

TysonRayJones:
Closing this PR (mostly for easy accounting after unitaryHACK), feel free to open another with a different example!

PR #787, opened 2026-06-10
Optimise bitwise extraction and insertion

### LINKED ISSUE #717: Optimise `getValueOfBits` and `insertBits`

lijunbie:
## Summary

- Add guarded BMI2 fast paths for `insertBits()` and `getValueOfBits()` using `_pdep_u64` / `_pext_u64` on x86-64 builds compiled with `__BMI2__`.
- Keep the existing portable loop fallback for non-BMI2 builds, CUDA/HIP compilation, small fixed-size cases, and `getValueOfBits()` calls whose bit indices are not in increasing order.
- Add unit coverage for `insertBits()` and `getValueOfBits()` over empty, low-bit, multi-bit, and non-sorted extraction cases.

Closes #717

## Validation

- Ran `git diff --check`.
- Ran an independent 64-bit equivalence check over 261 sample cases comparing the BMI2-style algorithms against the existing loop semantics.
- I could not run the full Catch2/CMake test suite locally because `cmake` is not installed in this environment.

### DIFF
--- a/quest/src/core/bitwise.hpp
+++ b/quest/src/core/bitwise.hpp
@@ -14,6 +14,11 @@
   #include <intrin.h>
 #endif
 
+#if defined(__BMI2__) && (defined(__x86_64__) || defined(_M_X64)) && !defined(__NVCC__) && !defined(__HIP__)
+  #include <immintrin.h>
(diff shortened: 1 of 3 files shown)


TysonRayJones:
That's some truly low effort AI slop

PR #789, opened 2026-06-11
test: add BMI2 bitwise benchmark baseline

tzh476:
This is the baseline PR requested in https://github.com/QuEST-Kit/QuEST/pull/779#issuecomment-2948379234.\n\nIt adds the same automated benchmark from #779, but is based on current devel and intentionally does not include the BMI2 helper implementation changes. The goal is to let CI run the identical benchmark against the current fallback/original bitwise helper paths for a like-for-like comparison with #779.\n\nValidation performed locally:\n- Confirmed the benchmark source matches #779 byte-for-byte.\n- Confirmed the CMake patch matches the benchmark-only commit from #779 via stable patch-id.\n- Ran `git diff --check`.\n\nI could not run the CMake build locally because this machine does not currently have `cmake` installed; CI should provide the cross-platform benchmark output.

### DIFF
--- a/examples/automated/CMakeLists.txt
+++ b/examples/automated/CMakeLists.txt
@@ -1,3 +1,9 @@
 # @author Tyson Jones
 
 add_all_local_examples()
+
+include(CheckCXXCompilerFlag)
+check_cxx_compiler_flag("-mbmi2" QUEST_COMPILER_SUPPORTS_MBMI2)
+if (QUEST_COMPILER_SUPPORTS_MBMI2)
+  target_compile_options(benchmark_bmi2_bitwise_cpp PRIVATE -mbmi2)
+endif()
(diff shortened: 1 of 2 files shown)


tzh476:
The baseline workflows are currently in `action_required` state (audit, compile, and test-free). Could you approve/run them when convenient so we can collect the fallback benchmark numbers for comparison with #779? Thanks!

tzh476:
For visibility: the upstream workflows for this baseline PR still appear to need maintainer approval, but I ran the same benchmark from this branch in a temporary fork-only workflow to get provisional fallback numbers:

- https://github.com/tzh476/QuEST/actions/runs/27348202274/attempts/3

That run used an AMD EPYC 7763 runner with Ubuntu 24.04 / GCC 13.3.0 and printed `BMI2 intrinsics: disabled`, as expected for the baseline/no-helper branch. I posted the provisional same-CPU comparison against #779 here: https://github.com/QuEST-Kit/QuEST/pull/779#issuecomment-4680842858

The authoritative signal should still come from this PR's upstream CI once approved.

tzh476:
Closing this helper baseline PR to stay within the unitaryHACK26 4 open PR rule. The primary optimization PR remains open for review.

PR #790, opened 2026-06-11
Fuse distributed prefix-suffix multi-SWAPs into one all-to-all round (GPU + controlled)

### LINKED ISSUE #595: Implement SWAP fusion

thedaemon-wizard:
# Fuse distributed prefix↔suffix multi-SWAPs into a single all-to-all round (GPU + controlled)

Closes #595

## Summary

QuEST relocates prefix (global / inter-node) qubits into the suffix (local)
region before applying a many-target dense matrix (`applyCompMatr`,
`applyMultiQubitMatr`) or a partial trace on a **distributed** register. This
"cache blocking" is performed by the internal
`anyCtrlMultiSwapBetweenPrefixAndSuffix`, whose own `@todo` asks for the
`k` disjoint prefix↔suffix SWAPs to be **fused**: historically they ran one at a
time, so an amplitude relocated to another node by one SWAP is relocated *again*
by the next — `k` communication rounds and up to `k/2·N` amplitudes moved.

This PR replaces that loop with a single **personalized all-to-all** exchange
over the `2ᵏ`-rank subcube, sending each amplitude **directly to its final node
in one round** (cf. cuStateVec distributed index-bit swap, mpiQulacs
[arXiv:2203.16044], [arXiv:2311.01512] §IV C–D, [arXiv:quant-ph/0608239]).

It is deliberately scoped to **complement, not duplicate**, the two excellent
CPU-only PRs already open against this issue:

- **#786** (@nathandelcid) — fuses the *uncontrolled* CPU multi-swap into ≤2
  waves within the existing `cpuCommBuffer`. (Maintainer: *"This is a beautiful
  diff! 🎉"*)
- **#785** (@zkasuran) — single-round-per-subset CPU/OMP fusion; its author
  explicitly defers the **async single round** and **GPU** kernels as follow-ups
  *"because I have no CUDA hardware to compile/test them."*

**This PR contributes exactly the pieces those two leave open**, fully compiled
and tested on real hardware:

1. **GPU (CUDA) pack/unpack kernels** for the fused swap — *compiled and tested
   on an NVIDIA Blackwell sm_120 GPU (CUDA 13.0)*, the precise piece #785 could
   not build.
2. **The controlled path** — fusion is applied with arbitrary control qubits /
   states intact (both competitors keep the controlled case on the slow
   per-swap path).
3. **A true single all-to-all round** (batched `Isend/Irecv` + one `MPI_Waitall`),
   with a persistent, lazily-grown staging workspace mirroring QuEST's existing
   `gpuCache` and the workspaces of cuStateVec / mpiQulacs.

If the maintainers prefer #786 for the CPU core, the GPU kernels and the
controlled-path handling here can be rebased on top of it.

## What changed

`anyCtrlMultiSwapBetweenPrefixAndSuffix` now collects the disjoint
prefix↔suffix pairs and:

- `k = 0` → returns; `k = 1` → delegates to the existing single-swap routine
  (**zero behavioural change**, no staging buffer);
- `k ≥ 2` → calls the new `fusedMultiSwapBetweenPrefixAndSuffix`.

### The fused algorithm (sub-block transpose over a `2ᵏ` subcube)
Ranks differing from ours only in the swapped prefix qubits' rank-bits form a
`2ᵏ`-rank subcube. Label a rank by its `k`-bit address `a`. A local amplitude
`j` with swapped-suffix pattern `v` maps to **rank `v`, new suffix pattern `a`**:

- the block with `v == a` (own address) **stays put, untouched** — no work;
- for each of the `2ᵏ−1` partner ranks, the local amplitudes whose suffix
  pattern equals the partner's address are packed, sent, and the received block
  is written back into the *same* local indices;
- the `2ᵏ−1` index sets are **disjoint**, so all exchanges run **concurrently in
  one round**.

Total moved `= (1 − 2⁻ᵏ)·N` (vs sequential `k/2·N`); peak extra memory is one
staging buffer `≤ N` per node, reused across calls and freed at
`finalizeQuESTEnv`.

### Files
| File | Change |
|------|--------|
| `core/localiser.cpp` | rewrite entry point; add `fusedMultiSwapBetweenPrefixAndSuffix`; retain `multiSwapSequentially` as reference + `QUEST_DISABLE_SWAP_FUSION` benchmark toggle |
| `comm/comm_routines.{cpp,hpp}` | `comm_exchangeAmpsToBuffersForFusedSwap` (CPU / GPU-direct / GPU-staged) + `exchangeArraysWithMultiplePartners` (single `Isend/Irecv` batch + one `MPI_Waitall`) |
| `cpu/cpu_subroutines.{cpp,hpp}` | OpenMP pack/unpack (templated on #targs) |
| `gpu/gpu_subroutines.{cpp,hpp}` | **CUDA pack/unpack** reusing the existing, tested `kernel_statevec_packAmpsIntoBuffer` / `kernel_statevec_anyCtrlSwap_subB` |
| `core/accelerator.{cpp,hpp}` | CPU/GPU dispatch + persistent staging-workspace cache |
| `api/environment.cpp` | free the staging workspace at teardown |

No public API change; the four existing call sites are untouched.

## Correctness

Disjoint single-pair swaps commute (already relied upon here), so the fused
result is identical to the sequential one. The `QUEST_DISABLE_SWAP_FUSION=1`
toggle runs the old sequential path for direct comparison. Density-matrix
callers (partial trace) pass ket+bra targets exactly as before; controls pass
straight through.

Verified against QuEST's deployment-independent brute-force reference
(`tests/utils`), which `QUEST_TEST_TRY_ALL_DEPLOYMENTS` runs across
serial / OMP / GPU / MPI:

| Deployment | Filter | Result |
|------------|--------|--------|
| Serial + OpenMP | `*PartialTrace*,*CompMatr*` | pass |
| **CUDA sm_120** | `calcPartialTrace,applyCompMatr2,applyCompMatr` | **All passed (20 780 assertions)** |
| CPU + MPI, np = 2 | `calcPartialTrace,applyCompMatr2,applyCompMatr,applySwap` | All passed (7 647) |
| CPU + MPI, np = 4 (fuses k = 2) | same | All passed (3 333) |
| CPU + MPI, np = 8 (fuses k = 3) | same | All passed (1 179) |
| CPU + MPI + OMP, np = 4, 3 threads | same | All passed (6 637) |
| **fused vs sequential toggle**, np = 4 | `calcPartialTrace,applyCompMatr2` | both All passed (1 148) |
| **GPU + MPI** | — | compiles cleanly |

## Performance

Micro-benchmark (`benchmarks/`): `applyCompMatr` (identity, to isolate
movement) on a forced-distributed 24-qubit statevector; `k = log2(np)` global
qubits are swapped in each call. Intel i5-13600K, OpenMPI 4.1.1, shared-memory
transport:

| np | k | fused µs/call | sequential µs/call | Δ | data moved fused / seq |
|----|---|---------------|--------------------|---|------------------------|
| 4  | 2 | 349 204 | 323 770 | +7.9 % | 0.75 N / 1.0 N |
| 8  | 3 | 377 673 | 377 410 | ~0 %  | 0.875 N / 1.5 N |
| 16 | 4 | 305 346 | 343 795 | **−11.2 %** | 0.94 N / 2.0 N |

The sequential cost grows linearly in `k`; fused stays at **one round** and its
moved-data saturates at `N`. Even on a single node — where the round-count
(latency) advantage is **entirely absent** because a "round" is a local memcpy —
fused already wins once `k ≥ 3–4`. On real multi-node interconnects, the
per-round network-latency savings make fusion advantageous at every `k`
(consistent with the cited literature). `benchmarks/README.md` analyses this in
full.

## Notes for reviewers
- `k ≤ 1` is byte-for-byte the old path; the staging buffer and all new comm
  only engage for `k ≥ 2` distributed registers.
- The GPU fused **comm-staging** path is compiled & the CUDA kernels are
  numerically tested on sm_120; multi-rank GPU runtime is limited here to one
  physical GPU, so the multi-node *GPU* path is compile-verified while the
  multi-node *CPU* path is runtime-verified at np = 1/2/4/8.
- The disabled meta-control optimisation at the old call site is left disabled.

## AI-assistance disclosure

Per the unitaryHACK [AI policy](https://unitaryhack.dev/ai-guide/): I used an AI
coding assistant (Anthropic Claude) to help draft the fused-swap algorithm
derivation, scaffold the pack/unpack kernels by mirroring QuEST's existing
`kernel_statevec_packAmpsIntoBuffer` / `kernel_statevec_anyCtrlSwap_subB`, and
to help write this description and the benchmark harness. I then **manually
reviewed, verified, and tested** every change myself:

- I can explain each modified routine and the sub-block-transpose derivation in a
  live review.
- All changes were compiled and run on real hardware — CPU, OpenMP, **CUDA
  sm_120 (Blackwell, CUDA 13.0)**, and CPU+MPI at np = 1/2/4/8 — and validated
  against QuEST's deployment-independent brute-force reference (the pass counts
  in the *Correctness* table above are my own runs).
- The fused-vs-sequential equivalence and the performance numbers were produced
  by me on the machine described in the benchmark section, not generated text.

No part of this PR was submitted unverified straight from an LLM.

### DIFF
--- a/quest/src/api/environment.cpp
+++ b/quest/src/api/environment.cpp
@@ -17,6 +17,7 @@
 #include "quest/src/core/autodeployer.hpp"
 #include "quest/src/core/validation.hpp"
 #include "quest/src/core/randomiser.hpp"
+#include "quest/src/core/accelerator.hpp"
 #include "quest/src/comm/comm_config.hpp"
 #include "quest/src/cpu/cpu_config.hpp"
 #include "quest/src/gpu/gpu_config.hpp"
@@ -458,6 +459,9 @@ void finalizeQuESTEnv() {
     // NOTE:
     // calling this will not automatically
     // free the memory of existing Quregs
+
+    // free the persistent fused-multi-swap staging workspace (host or device), if any
+    accel_clearFusedSwapSendCache();
 
     if (global_envPtr->isGpuAccelerated)
         gpu_clearCache(); // syncs first
--- a/quest/src/comm/comm_routines.cpp
+++ b/quest/src/comm/comm_routines.cpp
@@ -242,6 +242,45 @@ void exchangeArrays(qcomp* send, qcomp* recv, qindex numElems, int pairRank) {
 }
 
 
+void exchangeArraysWithMultiplePartners(
+    qcomp* send, qcomp* recv,
+    const vector<int>& partnerRanks,
+    const vector<qindex>& sendInds, const vector<qindex>& recvInds,
+    qindex numAmpsPerBlock
+) {
+#if QUEST_COMPILE_MPI
+
+    MPI_Comm mpiComm = comm_getMpiComm();
+
+    // each per-partner block is divided into power-of-2 messages; we issue ALL of every
+    // partner's asynchronous send/receives up-front (a personalised all-to-all) and wait
+    // once, so that a fused multi-swap completes in a single communication round
+    auto [messageSize, numMessages] = dividePow2PayloadIntoMessages(numAmpsPerBlock);
+
+    vector<MPI_Request> requests(2 * partnerRanks.size() * numMessages, MPI_REQUEST_NULL);
+    size_t r = 0;
+
+    for (size_t b=0; b<partnerRanks.size(); b++) {
+        int pairRank = partnerRanks[b];
+
+        // messages to/from distinct partners share tags safely (distinguished by rank);
+        // within a partner, unique tags permit out-of-order arrival (UCX adaptive-routing)
+        for (qindex m=0; m<numMessages; m++) {
+            int tag = static_cast<int>(m);
+            MPI_Irecv(&recv[recvInds[b] + m*messageSize], messageSize, MPI_QCOMP, pairRank, tag, mpiComm, &requests[r++]);
+            MPI_Isend(&send[sendInds[b] + m*messageSize], messageSize, MPI_QCOMP, pairRank, tag, mpiComm, &requests[r++]);
+        }
+    }
+
+    // single wait completes the whole round (MPI will automatically free the request memory)
+    MPI_Waitall(requests.size(), requests.data(), MPI_STATUSES_IGNORE);
+
+#else
+    error_commButEnvNotDistributed();
+#endif
+}
+
+
 
 /*
  * PRIVATE ASYNC SEND AND RECEIVE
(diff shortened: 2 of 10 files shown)


TysonRayJones:
Hi Amon,

This is some great work, although we can only nominate one recipient of the UnitaryHACK challenge, which has gone to the author of the original diff. Your extension to support control qubits, and the GPU implementation, is a great help however - we would gladly also integrate this (after some cleanup) into the QuEST source, as an external contribution.

After #785 is merged into `devel`, it will be easier to see your original diff in this PR. We'll run some benchmarks and verification on our end, so there's nothing further to do for now (ignore the failing CI, grr!). Thanks for the great work!

thedaemon-wizard:
Thanks @TysonRayJones, that's completely understood — congrats to @zkasuran on the challenge. I'm very happy to have the control-qubit and GPU fused-multi-SWAP integrated as an external contribution.

Whenever #785 lands on `devel` I'll rebase this PR onto it so only my incremental diff (controlled path + CUDA pack/unpack + persistent staging buffer) remains, and I'll do the cleanup pass then — including dropping the `QUEST_DISABLE_SWAP_FUSION` benchmark toggle, which was only there for the fused-vs-sequential measurements. Just let me know if you'd prefer any particular structure.

The Linux CI (CPU/OMP/CUDA/MPI) is green; the failures are the Windows CUDA jobs. Thanks again!

PR #797, opened 2026-06-15
ci: locate Visual Studio with vswhere in the MSVC CUDA setup

zkasuran:
## What

The Windows + CUDA CI jobs fail in the "Setup CUDA (MSVC)" step. `setup_cuda.ps1` copies the CUDA MSBuild integration into Visual Studio's `BuildCustomizations` folder, found by globbing a hard-coded `C:\Program Files\Microsoft Visual Studio\2022\*` path. That path no longer exists on the current `windows-latest` runner image, so the step exits 1 with "Cannot find path 'C:\Program Files\Microsoft Visual Studio\2022' because it does not exist" before anything compiles.

This is hitting the whole repo's Windows CUDA matrix right now (for example #785 and #791), not one branch.

## Fix

Resolve the VS install path with `vswhere`, the supported version-independent way to locate Visual Studio, then glob `MSBuild\Microsoft\VC\*\BuildCustomizations` under it. `vswhere` ships at a stable location on every GitHub Windows image, so this keeps working across runner image bumps.

One file, the MSVC CUDA setup script. It only runs for the Windows MSVC CUDA matrix, so it cannot affect the Linux, macOS or non-CUDA builds.

I do not have a Windows runner locally, so the proof is this PR's own Windows CUDA matrix. If the vswhere call needs adjusting I will iterate here.

AI disclosure: drafted with help from Claude (Anthropic). I traced the failure to the "Setup CUDA (MSVC)" step (the VS2022 path drift) across the failing jobs and an unrelated PR before writing the change. Verified by CI.

### DIFF
--- a/.github/scripts/setup_cuda.ps1
+++ b/.github/scripts/setup_cuda.ps1
@@ -83,7 +83,15 @@ foreach ($dir in $directories) {
 
 # Add msbuild cuda extensions
 $msBuildExtensions = (Get-ChildItem  "$src\visual_studio_integration\CUDAVisualStudioIntegration\extras\visual_studio_integration\MSBuildExtensions").fullname
-(Get-ChildItem 'C:\Program Files\Microsoft Visual Studio\2022\*\MSBuild\Microsoft\VC\*\BuildCustomizations').FullName | ForEach-Object { 
+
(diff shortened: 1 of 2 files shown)


TysonRayJones:
Alternate patch has reached `devel`

PR #808, opened 2026-06-29
Add ad-hoc edge-case tests to CI

TysonRayJones:


TysonRayJones:
Welp, the 12-qubit statevector unit tests are way too slow to run, even when setting all environment variables to their least intensive. I believe this is because the number of values of `numCtrls` and `numTargs` is still very large - though we are really only interested in the settings where `numCtrls = numTargs = 6` (and perhaps also where one is `<=5` (activating an optimisation) and the other is `6` (no optimisation). We may have to target this specific regime in bespoke `TEST_CASE`, if possible without duplicating all the `operator` unit test harness. Bah!

TysonRayJones:
I've been brainfarting; given the reference _state_ 12-qubit density matrix is intractable (becoming a 24-qubit 256 MiB statevector), so too is the 12-qubit _reference operator_ being attemptedly prepared for multiplication upon the 12-qubit statevector. Regardless of the ctrls + targs permutations, `getFullStateOperator()` hangs on a `12`-qubit state.

PR #818, opened 2026-08-18
Abort undersized auto-distributed Qureg instead of replicating it

iarejula-bsc:
In a multi-node environment, `createQureg()` can currently return a Qureg which is replicated across every node rather than distributed between them. Every node then redundantly simulates the same state, so all but one of the allocated nodes do no useful work, and nothing warns the user. The auto-deployer reaches that outcome by two routes:

- the state has too few amplitudes to divide evenly between the nodes (`numQubits < log2(numNodes)`), so it cannot be distributed at all
- the state would divide evenly, but distribution is judged not worthwhile, because each node would be left with fewer than `MIN_NUM_LOCAL_QUBITS_FOR_AUTO_QUREG_DISTRIBUTION` (26) local qubits.

The solution provided is to require that an automatic deployment utilises every node of the environment, and to abort with a clear error otherwise, the same way already done in validation (e.g. the "node count must be a power of 2" check), so callers can react instead of QuEST silently proceeding. Each route reports its own reason: how many qubits would be needed to divide the state between the nodes, or how many amplitudes each node would be left holding.

Explicitly requesting a non-distributed Qureg (via `createCustomQureg()` or `createQuregFromEnvDeployments()`) remains permitted, since that replication is deliberate; only deployments left to the auto-deployer are affected.

`reportQuESTEnv()` now reports the sizes which can no longer be automatically deployed, rather than advertising a deployment which `createQureg()` would reject.

I would like to discuss the use of *Malleability instead of aborting*.  Where the runtime supports malleability (DMR library), a better response is available, shrink the job to a node count at which the Qureg does distribute, and carry on,  which would turn this error into a resize and leave the abort as the fallback for whatever malleability cannot cover. This would not affect the core of Quest and would be low intrusive, as it would only modify the autodeployer, which i think it fits the use of malleability.

Disclosure: this contribution was developed with support of Opus.

### DIFF
--- a/quest/src/api/environment.cpp
+++ b/quest/src/api/environment.cpp
@@ -340,6 +340,9 @@ void printQuregAutoDeployments(bool isDensMatr) {
     prevGpuAccel = 0;
     prevMulti    = 0;
 
+    // assume 1 qubit is deployable, so that an undeployable first row is still reported
+    bool prevCanDeploy = true;
+
     // test to theoretically max #qubits, surpassing max that can fit in RAM and GPUs, because
     // auto-deploy will still try to deploy there to (then subsequent validation will fail)
     int maxQubits = mem_getMaxNumQuregQubitsBeforeGlobalMemSizeofOverflow(isDensMatr, globalEnvPtr->numNodes);
@@ -350,22 +353,34 @@ void printQuregAutoDeployments(bool isDensMatr) {
         useDistrib  = modeflag::USE_AUTO;
         useGpuAccel = modeflag::USE_AUTO;
         useMulti    = modeflag::USE_AUTO;;
-        autodep_chooseQuregDeployment(numQubits, isDensMatr, useDistrib, useGpuAccel, useMulti, *globalEnvPtr);
+        // mustUtiliseAllNodes=false: this merely queries sizes, so must never abort upon
+        // those which createQureg() would reject; we report them as undeployable below
+        autodep_chooseQuregDeployment(numQubits, isDensMatr, useDistrib, useGpuAccel, useMulti, *globalEnvPtr, false, __func__);
 
-        // skip if deployments are unchanged
-        if (useDistrib  == prevDistrib  &&
+        // createQureg() rejects auto-deployments which would replicate the Qureg between nodes
+        bool canDeploy = (globalEnvPtr->numNodes == 1) || useDistrib;
+
+        // skip if deployments are unchanged, or remain unavailable (the other deployments
+        // still vary while unavailable, but are not reported, so must not open a new row)
+        if (!canDeploy && !prevCanDeploy)
+            continue;
+        if (canDeploy == prevCanDeploy &&
+            useDistrib  == prevDistrib  &&
             useGpuAccel == prevGpuAccel &&
             useMulti    == prevMulti)
             continue; 
 
         // else prepare string summarising the new deployments (trailing space is fine)
-        string value = "";
-        if (useMulti)
-            value += "[omp] "; // ordered by #qubits to attempt consistent printed columns
-        if (useGpuAccel)
-            value += "[gpu] ";
-        if (useDistrib)
-            value += "[mpi] ";
+        string value = "(no automatic deployment available)";
+        if (canDeploy) {
+            value = "";
+            if (useMulti)
+                value += "[omp] "; // ordered by #qubits to attempt consistent printed columns
+            if (useGpuAccel)
+                value += "[gpu] ";
+            if (useDistrib)
+                value += "[mpi] ";
+        }
 
         // log the #qubits of the deployment change
         rows.push_back({printer_toStr(numQubits) + " qubits", value});
@@ -374,6 +389,7 @@ void printQuregAutoDeployments(bool isDensMatr) {
         prevDistrib  = useDistrib;
         prevGpuAccel = useGpuAccel;
         prevMulti    = useMulti;
+        prevCanDeploy = canDeploy;
     }
 
     // tailor table title to type of Qureg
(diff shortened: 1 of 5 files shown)


TysonRayJones:
Could you elaborate in words the decision made? It sounds like you propose to remove the automatic disabling of distribution. The existing functionality is very useful to ensure that the chosen parallelisations/deployments are _performant_ for the user's `Qureg`. Forcing distribution, when the environment is distributed, will pointlessly cripple performance. Keep in mind that a user may be creating _multiple_ `Qureg` of varying sizes within the same execution, and it is senseless to force the small ones to distribute their data. Please see instead `createForcedQureg`

iarejula-bsc:
I am not removing the distribution, but the replication (same qureg on all nodes).

This comes because I have been doing benchmarks on Quest for the last months. I realized that after the 26 local qubits limit, the autodeployer just wastes resources. As you can see in the image, with 32 nodes it uses all resources but only 1 node is useful, as the rest are replicas with the same result. Personally, I do not see why this is worth it, nor was I aware of this behavior until I saw the benchmarks and analyzed the code to understand the issue of 32 nodes being the worst performant (which is not a good ux, cuz if i did not run the benchmark i wouldnt be aware of 32 is worse than 16 and if i did not read the code i would not know the reason of this behaviour)

The idea of this PR is to notify the user and avoid wasting resources, but still allow the user to use this functionality in the "no auto" mode, if they really want it.
results_Draper_adder_sizes

I understand that this limits the scenario where a user allocates 16 nodes for a circuit of 30 qubits and another of 5 qubits: the first one distributes, and the second one fails. I believe that not wasting resources is the better default, with wasting them being opt-in, I am open to discussing it if you disagree and think it is worth to discuss it. 

However, I feel the best option here is to use malleability to reallocate the allocated nodes, but that is another discussion.

Edit: the optimized vs baseline is not related to this discussion. The only line that matters is baseline.

TysonRayJones:
Hi there,

There still seems to be a misunderstanding - the "wasting" of computation across processes, when a `Qureg` has been replicated across them, is deliberate. It is a _speedup_ in settings where some `Qureg` are distributed, and others are not - which is the expected and standard scenario when a user is deploying a multi-node simulation (otherwise, there is no reason a user should ever launch their <26 qubit job with multiple MPI processes; they are inviting a slowdown!). 

Perhaps you can see the speedup more clearly when it comes time to perform some two-`Qureg` computation between a distributed and non-distributed `Qureg`. The non-distributed `Qureg` must contain corresponding data in every process spanned by the distirbuted `Qureg`. Under your proposal of not-duplicating the non-distributed `Qureg`, then at that point in time, we must broadcast the non-distributed `Qureg` from some process, say `0`, to every other process. That broadcasting is superfluous communication, which is our main contribution to runtime in our typical network-bandwidth-bound settings. So we are pointlessly waiting for the single process responsible for the non-duplicated `Qureg` to broadcast its entire state, to every other node! That's a greater bottleneck then typical pair-wise communication between processes (QuEST's main cost during gate application), since at least there, all pairs of nodes send/recv independently of one another. But in this scenario, process `0` does _all_ the sending!

Consider instead the current scheme, where every node is _always_ simultaneously evolving the non-distributed `Qureg`. At the time of the two-`Qureg` computation with a distributed `Qureg`, no communication is necessary since every process already has all needed data, and the computation can proceed immediately.

The superfluous communication invited by not-duplicating non-distributed `Qureg` is actually still worse than this. QuEST boasts API agnosticism to the deployment mode; so decisions the user makes in their control flow (e.g. `if (getProb(smallQureg) > 1): applyX(bigQureg)`) should always work, even when distributed. But in your scheme, this logic will fail. The user would need to broadcast the result of `getProb(smallQureg)`, which is only known to process `0`, to all other processes which share `bigQureg`. Otherwise, the other processes do not know the output of `getProb()`, cannot determine the `if` branch faithfully, fall out of consensus/synchronisation, and the system will likely deadlock.

So, lots of additional, superfluous, slowing communication is necessary when trying to let "non-processes relax" during simulation of a non-distributed, non-duplicated `Qureg`. It's a strict slowdown. Hence, your benchmarking results are very suspicious to me. I suspect you are measuring a special case _without_ performing the additional communication needed of the general case (maybe you're simulating a single `Qureg` - the scenario where a distributed launch is totally pointless). That you see duplication causing a slowdown is _very_ suspicious; process `0` performing superfluous, embarrassingly parallel computation should _not_ slowdown independent process `1`, for example. Perhaps you are oversubscribing a single machine to pointlessly use distribution? This will, of course, not yield representative performance data

iarejula-bsc:
I now see the point of why replicating the qureg makes sense. However, this behavior is not well documented, and it can lead users to deploy on bigger clusters expecting a speedup, only to end up with the performance of a single node while wasting the extra nodes. This PR could be turned into a documentation PR instead. That said, I feel both behaviors could coexist, giving users the tools to either replicate or assert, depending on their use case. Since you are more experienced in this field, I'll leave the decision to the Quest team.

Regarding the benchmark, the only relevant part for this discussion is the baseline. The other results are our own optimizations, which are not part of this discussion, and I think including them was misleading. I just had this chart on hand and used it instead of generating one specifically for this discussion.

That said, the benchmark is correct. All of them were run on MareNostrum, using up to 64 real nodes, not a single machine as you mentioned. I'm also not sure what seems suspicious to you, the behavior is exactly as expected: walltime decreases as we go from 1 to N nodes, and once the limit is reached, it becomes embarrassingly parallel for the qureg. This is shown by 1 node, 32, and 64 having the same walltime, which is expected.

The scenario is a single Qureg running a circuit with gates gates that requires and not require communication, and the whole point of the distribution is to speed up the process, which is exactly what the data shows.

